PR #49, opened 2026-07-08
feat(web): spec-driven Welcome screen for first-touch

rsolmano:
## What

When no workspace is active (a fresh install, or after archiving the last workspace), the shell used to render the full IDE surface — empty center, an empty Specs/Files/Changes rail, and a disabled terminal. This replaces that dead, cluttered first impression with a branded, **spec-driven Welcome screen** beside the projects rail.

The framing is deliberate: ThinkRail isn't just a UI for `pi` — it makes **specs the source of truth**. The Welcome copy and the spec-aware "Set up project" flow lead with that.

## Behaviour

The shell branches its body on `activeWorkspaceId`:
- **Active workspace** → today's 3-column IDE (unchanged).
- **No active workspace** → the projects rail (still resizable) + a centered `WelcomePanel`.

`WelcomePanel` shows the ThinkRail wordmark, the active project's name as an eyebrow, a constant spec-first pitch, and **adaptive cards**:

| State | Cards |
| --- | --- |
| No projects | **Open project** |
| Project, has specs | **Start building** + Open project |
| Project, no specs | spec-first **Set up project** + Start building + Open project |

- **Set up project** opens the New-Workspace dialog with a **pre-filled** prompt that nudges the `project-setup` workflow (idea → `goal-and-requirements.md`).
- **Open project** hangs the shared add-project dropdown off the card (same menu as the projects-rail `+`).
- **Start building** is the intent-first framing of the create-and-kick-off flow (a worktree-isolated workspace + a chat); *workspace* is the mechanism, not the label.

## How it detects specs

`Project.hasSpecs` is computed by the host **per read** (repo-root `goal-and-requirements.md`), never persisted — so it always reflects disk. Additive, non-breaking wire field.

## Notable refactors

- Extracted the projects `+` menu into a shared **`AddProjectMenu`** (reused by the Welcome "Open project" card).
- Extracted **`projectActions`** (open-project orchestration) so `ProjectTree` and `WelcomePanel` don't duplicate it.
- `NewWorkspaceDialog` gained an optional `initialPrompt`.

## Specs

Updated as part of the change: `packages/contracts`, `packages/server/src/projects`, `apps/web/src/shell`, `apps/web/src/panels`, `apps/web`.

## Tests

`bun run e2e` — **21/21 green**. New `e2e/welcome.spec.ts` covers all three states (including a **clean install with no projects imported**), the "Open project" popover, and the pre-filled setup dialog. `shell.spec.ts` now asserts the Welcome screen on a fresh load. `lint` + `typecheck` pass.

## Screenshots

Thinkrail without any projects:
welcome-01-no-projects

Project without specs (Set up project):
welcome-03-needs-setup

Project with specs:
welcome-02-has-specs


apps/web/src/panels/WelcomePanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+						{openProjectCard({ subtitle: "Add another local git repository." })}
+					</>
+				)}
+			</div>
+
+			{dialog ? (
+				<NewWorkspaceDialog
+					open
+					projectId={dialog.projectId}
+					{...(dialog.prompt ? { initialPrompt: dialog.prompt } : {})}
```
danyaberezun:
```suggestion
					initialPrompt={dialog.prompt}
```
?

packages/server/src/spec/specs.ts
```diff
@@ -23,6 +23,30 @@ export function evictSpecIndex(workspaceId: string): void {
...
 
+/** One reused index per project root, for the project-level `hasSpecs` check below (revalidate-on-read). */
+const projectIndexes = new Map<string, SpecIndex>();
+
+/**
+ * Whether a project's repo root carries **any** registered spec (a file with `id` + `type` frontmatter,
+ * anywhere under the root) — the signal the Welcome screen uses for its "Set up project" suggestion.
+ * Uses the same derived, revalidate-on-read index as the agent's spec tools, so it's robust to any spec
+ * filename/casing (not just a lowercased `goal-and-requirements.md`) and always reflects the filesystem.
+ * A per-root index is reused so repeat reads (welcome, project.list) only pay the glob once. Defensive:
```
danyaberezun: Honestly speaking, `pay the glob once` is interesting wording. I do understand it refers to `glob` as util but maybe smth like `skip re-parsing unchanged ...bla-bla...` might be a bit more natural.

apps/web/src/panels/WelcomePanel.tsx
rsolmano: Good call — done in 99cd311. The bare `initialPrompt={dialog.prompt}` wouldn't compile as-is: `prompt` was `string | undefined` and the prop is `initialPrompt?: string`, which `exactOptionalPropertyTypes` rejects (that's why the conditional spread was there). Fixed by making the dialog's `prompt` always a string ("" for a plain create), so the prop can be passed directly.

packages/server/src/spec/specs.ts
rsolmano: Agreed — reworded in 99cd311 to: "A per-root index is reused across reads (welcome, project.list) so its parse cache skips re-reading unchanged spec files — only new or changed ones are re-parsed." (Also more accurate: the glob runs on every read; it's the *parse* that's cached.)

danyaberezun (approved):

PR #55, opened 2026-07-09
update .gitignore

danyaberezun:
## What

Housekeeping on `.gitignore`, plus stops tracking the JetBrains `.idea/` project files.

### `.gitignore`
- **Generated build modules → glob.** `apps/cli/scripts/build-binary.ts` writes two transient modules (`web-assets.generated.ts` **and** `bundled-extensions.generated.ts`) and deletes them in a `finally`. Only the first was ignored, so an interrupted build (Ctrl-C / crash before cleanup) could leave `bundled-extensions.generated.ts` showing in `git status`. Replaced the single line with `apps/cli/src/*.generated.ts` so any sibling is covered.
- **Local env / secrets.** Added `.env`, `.env.local`, `.env.*.local`. Vite auto-loads `.env*` and the app handles provider auth tokens — cheap insurance against committing secrets. Template files (`.env.example`) stay committable.
- **Editor/tooling dirs.** Ignore `.idea/` and `.claude/`.

### Stop tracking `.idea/`
Removed the 6 tracked `.idea/*` files (kept on disk via `git rm --cached`). They're IDE-regenerated scaffolding; the only non-default rule was a stale **Python/OpenAI** inspection — wrong for this Bun/TypeScript repo. Real shared config already lives in `biome.json`, `tsconfig*.json`, `turbo.json`.


.gitignore
```diff
@@ -11,6 +12,12 @@ e2e/screenshots/
 
 docs/*
 .playwright-mcp/
+.claude/
```
rsolmano, line 15: I think we may need it. For shared skills, etc. If there something claude related you don't wanna merge - just have it outside the project, in ~/.claude

rsolmano (approved):

.gitignore
danyaberezun: in case one need it, he/she can change .gitignore
right now, we do not use it at all

PR #65, opened 2026-07-13
feat(e2e): headless workflow-test harness + routing suite (pi-agent-driven)

SBOne-Kenobi:
## What this gives us

The workflow skills (`packages/pi-thinkrail-workflow`) are the most behavior-critical, least-testable layer we ship — until now their only verification was a manual live run per skill. This PR makes skill behavior **cheaply observable**: a headless harness drives a **real in-process pi agent** through the production wiring, and a workflow test is **one `defineScenario` call**. The routing layer — every row of both routers' classification tables — is now covered and green.

Three commits: harness (`70446be`) · suites (`c2f9f60`) · spec alignment (`ddbb3aa`). Durable design lives in `e2e/workflows/SPEC.md` (`module-workflow-tests`).

## What really matters

**1. The verdict model — an LLM never decides *pass*.**
- **Binding:** deterministic checks + signals over the captured event log and workspace — a skill load *is* the `read` of its `SKILL.md`; artifacts asserted on disk; forbidden tools/paths absent. A crashed run records `crashed` and can never claim a pass.
- **Input/diagnosis only:** persona responder + user simulator (cheap LLM plays the human; answers ride the production `answerQuestion` bridge), watchdog (budget tripwires + abort-with-diagnosis), rubric judge (**advisory** — warn + run log, never a test failure).

**2. Production wiring, no browser, no HTTP host.** Sessions come from the new `@thinkrail/server/agent` subpath export — a deliberate second entry that skips the Bun-only `host` under the node-run worker (recorded in `packages/server/SPEC.md`; web/CLI boundary rules unchanged). Isolated `PI_CODING_AGENT_DIR` via the shared e2e global setup; never `setModel`.

**3. Cost + safety by construction.** `stopWhen` aborts the moment the question is answered (routing scenarios ≈ 10s); an unscripted interview can never hang (dialog ladder: script → persona → deterministic fallback — a throwing script matcher degrades down the ladder instead of crashing the worker). On-demand only: `bun run test:workflows` — needs pi auth, spends real tokens, excluded from `test`/pre-commit/CI/`e2e` (main config `testIgnore`s `workflows/**`).

**4. Mid-flow entry**, so tests don't replay whole conversations: artifact presets (the task-spec is the workflow's spine) and recorded-transcript fixtures (`SessionManager.open`; born via `THINKRAIL_WORKFLOW_RECORD=1`; fixture markdown masked as `*.md.test` so fixture specs never leak into the host's own spec graph).

## Suites

- `routing.live.spec.ts` — **all six classifications green**: root router (feature → `brainstorming` · onboarding → setup family · question → no worker loads) and the dispatcher via the app's exact `/skill:` seed (empty → `starting-a-new-project` · code-only → `importing-a-codebase` · specced → review/extend offer, both workers forbidden).
- `smoke.live.spec.ts` — infra smokes: mid-flow `brainstorming` round-trip (preset + persona answer + simulator + decision on disk); transcript continuation from a recorded fixture.
- `harness.unit.spec.ts` — the pure parts (parsers, predicates, checks, tripwires, crash recording, fixture masking); no agent, no auth.

## Found a real bug (recorded, deliberately not fixed here)

**Pure questions bypass the root router.** `WORKFLOW_RULE` names "a request, feature, change, fix, or fresh project idea" — **questions aren't listed** — while `choosing-a-workflow`'s description claims them, so the agent answers without reading the router and its one-line "no workflow" declaration never happens. The bypass is question-specific (a non-question "anything else" request was observed taking the full path). Binding outcome stays correct, so the scenario passes and the **advisory judge re-flags the drift on every run** until the wording is reconciled — that's a change to the always-on, byte-stable rule and gets its own task. Recorded in `packages/pi-thinkrail-workflow/skills/SPEC.md` § Current limitations & gaps.

## Spec alignment (`ddbb3aa`)

Specs are this repo's ground truth, so the graph was reconciled with what shipped: each fact now has one home (`module-thinkrail-workflow` § Testing links the harness spec instead of restating it); the family table carries per-skill observation status (routing suite / smokes / 2026-07 manual runs — `asking-user-questions` and `writing-specs` now *observed by use*; `brainstorming`'s missing `writing-specs` reference confirmed as an observed defect); rule 14 (verify-by-use) stays suspended only for want of a durable run record (runs land in the gitignored `e2e/.workflow-runs.jsonl` by decision); and `architecture.md` / `goal-and-requirements.md` no longer read as contradicting the shipped skill system — "Workflows" out-of-scope is now scoped to the workflow *product layer*.

## Verification

- `bun run test:workflows` — **14/14** live against a real provider (pre-squash; the squash is tree-identical — `git diff backup HEAD` empty)
- `harness.unit.spec.ts` — 10/10 re-run at the final tip
- `bun run e2e` — 28/28 (browser suite unaffected)
- `lint` / `typecheck` (8/8 packages) / `spec_validate` — green at the final tip

## Follow-up (scenario definitions only, no new machinery)

- Slice 3: worker flows end-to-end (`starting-a-new-project` interview via the user simulator, `importing-a-codebase`, full `brainstorming`) — tracked in `packages/pi-thinkrail-workflow/SPEC.md` § Testing.
- Reconcile `WORKFLOW_RULE` vs the router's "question" claim (the recorded limitation; own brainstormed task).


SBOne-Kenobi:
Follow-up improvements (slice 3 worker flows, eval-style harness enhancements, re-arming verify-by-use) are now tracked in #67.

danyaberezun (dismissed):

rsolmano (approved):

PR #71, opened 2026-07-14
feat(web): toasts for wire failures with no chat to land in

rsolmano:
> **Stacked on `auth-provider-status-ui`** — base that PR, review this one on top. The diff below is only the toast pass.

## What & why

The auth branch left two `// A toast comes with the error-handling pass.` TODOs, and a few sibling sites that swallow errors or (worse) leave an unhandled rejection. This introduces a small toast system and wires up **all** of them — the home the comments pointed at.

## The system

- **store** (`appStore`): a `toasts` queue + `pushToast`/`dismissToast`, and an ergonomic `toast.error/success/info(message, title?)` helper so a non-React `.catch` can fire one. It's the place a **rejected wire call with no chat tab to host an error turn** lands — complementing the existing `appendErrorTurn` (which handles the in-chat case).
- **components/ui/toast**: Radix Toast primitives, themed with our tokens (a left accent bar carries severity — red/green/primary). Presentational only.
- **panels/Toaster**: composes the primitives against the store; the **shell mounts it once** (over both the workspace and welcome layouts). Radix owns the auto-timeout + swipe/hover-pause; every close routes back through `dismissToast` so the store stays the single source of truth. **Errors persist until dismissed; success/info time out (~5s).**

## Sites wired

| Site | Before | After |
|---|---|---|
| `NewWorkspaceDialog` — `workspace.create` fails | dialog silently refuses to close | toast (dialog stays open to retry) |
| `NewWorkspaceDialog` — `session.create` fails | `console.error` only — chat kick-off dropped invisibly | toast |
| `ProvidersSettings` — OAuth `loginStart` fails | silent | toast |
| `ProvidersSettings` — `loginReply` fails | `.catch(() => {})` — login stranded | toast |
| `ProvidersSettings` — `setApiKey` / `logout` reject | **unhandled rejection** — card went quiet | `catch` + toast |

## Verification

- Store slice unit-tested (append / dismiss / no-op-on-unknown-id / helper variants).
- Full **no-agent e2e suite green** (37/37), including the API-key round-trip that now exercises the new `catch` blocks.
- `lint` + `typecheck` pass.
- Toast render + variant-based timeout confirmed visually (error persists, success auto-dismisses).

No dedicated toast e2e: the no-agent suite has no deterministic server-side failure hook to trigger one without a test-only backdoor, so the store contract is covered by unit tests instead.

## SPECs

Updated `store`, `components/ui`, `panels`, and `shell` SPECs (spec-leads-the-code).

## Screenshots

toasts-all-variants
toasts-closeup


apps/web/src/panels/ProvidersSettings.tsx
```diff
@@ -75,6 +76,10 @@ export function ProvidersSettings() {
 			try {
 				await getTransport().request("provider.setApiKey", { providerId, key });
 				await load();
+			} catch (err) {
+				// A rejected save (malformed key, disk write failed) was previously an unhandled rejection — the
+				// key field just went quiet. Surface it so the user knows the key didn't take.
+				toast.error(errorText(err), "Couldn't save the API key");
 			} finally {
 				setBusyProvider(null);
 			}
```
danyaberezun:
```suggestion
			} catch (err) {
				// A rejected save (malformed key, disk write failed) was previously an unhandled rejection — the
				// key field just went quiet. Surface it so the user knows the key didn't take.
				toast.error(errorText(err), "Couldn't save the API key");
				return;
			} finally {
			    setBusyProvider(null);
			}
			await load(); // has own error banner
```

apps/web/src/panels/ProvidersSettings.tsx
```diff
@@ -88,6 +93,9 @@ export function ProvidersSettings() {
 			try {
 				await getTransport().request("provider.logout", { providerId });
 				await load();
```
danyaberezun: same

danyaberezun:
One site of the same failure class was missed: `startChat` in `apps/web/src/panels/CenterTabs.tsx` (still `// Ignored until the error-handling pass.`) — a failed `session.create` makes "+ New chat" do nothing, silently. Three lines with this PR's machinery: `catch (err) { toast.error(errorText(err), "Couldn't start the chat"); }`

apps/web/src/store/appStore.ts
```diff
@@ -856,4 +873,26 @@ export const useAppStore = create<AppState>((set) => ({
 	closeSettings: () => set({ settingsOpen: false }),
 	setSettingsSection: (section) => set({ settingsSection: section }),
 	requestChangesView: (workspaceId, path) => set({ changesRequest: { workspaceId, path } }),
+	pushToast: (toast) => {
+		const id = crypto.randomUUID();
+		set((s) => ({ toasts: [...s.toasts, { ...toast, id }] }));
+		return id;
+	},
```
danyaberezun, line 890: Errors persist until dismissed and there's no dedup/cap — retrying the same failure stacks identical cards, and a burst can grow past the viewport (`max-h-screen`, no scroll), hiding the newest. Minimal guard:

```suggestion
	pushToast: (toast) => {
		let id = "";
		set((s) => {
			// Coalesce an identical toast (a retried failure) instead of stacking a twin.
			const twin = s.toasts.find(
				(t) => t.variant === toast.variant && t.title === toast.title && t.message === toast.message,
			);
			if (twin) {
				id = twin.id;
				return {};
			}
			id = crypto.randomUUID();
			// Cap the queue — beyond a screenful the oldest drop.
			return { toasts: [...s.toasts, { ...toast, id }].slice(-5) };
		});
		return id;
	},
```

apps/web/src/panels/ProvidersSettings.tsx
rsolmano: Applied in 630a2ad (with tabs — the suggestion had a space-indented `finally`). Worth noting the mislabeling couldn't actually fire today: `load()` never rejects — it catches internally and renders the pane's error banner. But that contract was invisible and load-bearing, so the restructure is right: the catch's scope now matches its message structurally instead of relying on `load()`'s internals, and `load()` got a doc comment pinning the never-rejects contract.

apps/web/src/panels/ProvidersSettings.tsx
rsolmano: Applied the same shape here (630a2ad). One accepted trade-off: busy now clears before the status re-read, so Sign out re-enables for the one-round-trip window while the card is still rendered — a double-click there would just surface a "Couldn't sign out" toast, so not worth extra machinery.

apps/web/src/store/appStore.ts
rsolmano: Confirmed (the viewport is `max-h-screen` with no scroll, so a burst really does clip) and applied in 630a2ad. Two deviations from the snippet: implemented with zustand's `get()` instead of the closure-id-inside-`set` dance (same semantics, reads cleaner), and the cap is a named `MAX_TOASTS = 5` with the why (oldest drop keeps the newest visible). Known limitation: a coalesced success/info twin doesn't restart the live toast's 5s timer — fine for the real target, persistent error toasts. Store SPEC updated; both behaviors pinned by unit tests (coalesce returns the twin id + dismiss-then-repush gets a fresh one; cap drops the oldest).

rsolmano:
Good catch — wired in 630a2ad: `startChat` now toasts "Couldn't start the chat". Also wired `onReopenChat` right above it, which is the same failure class (a history-chat click whose `session.getMessages` rejection silently did nothing — it still stays in history for a retry, but now says why). Swept the rest of `apps/web` for the class: the remaining `.catch(() => {})` sites are deliberate fire-and-forget cleanups (terminal detach, login cancel), so this closes it out. Panels SPEC updated with both sites.

danyaberezun (approved):

PR #82, opened 2026-07-16
feat: gitignored .thinkrail/context/ for ephemeral workspace task files

rsolmano:
## What

Adds a **per-workspace scratch dir** for files needed only during task execution that must **never** be committed: the worktree-local, gitignored `.thinkrail/context/`. Skills that produce temp docs (brainstorming's `task-spec`s, and any future working files) write them there instead of into the committed tree.

## Why it works

The spec store scans the whole worktree ignoring only `node_modules/.git/dist/build` — **not** `.gitignore`. So a task-spec under `.thinkrail/context/` stays fully discoverable by `spec_grep`/`spec_get`/`spec_graph` and the Specs viewer, while being invisible to git. Ephemeral for git, live for the agent.

The `.gitignore` is a single `*`, which matches the `.gitignore` file itself → the whole dir has **zero git footprint** (nothing in `git status`, nothing accidentally committable). Verified.

## Changes

- **Host** — `createWorkspace` seeds `.thinkrail/context/.gitignore` (`*`) in every new worktree. (`packages/server/src/workspaces/` + SPEC + test)
- **`projectHasSpecs`** now counts only **durable** specs (excludes `task-spec`), so an ephemeral scratch doc never flips the Welcome screen's "Set up project" signal. (`packages/server/src/spec/` + SPEC + test)
- **Workflow skills** — `brainstorming` writes task-specs to `.thinkrail/context/TASK-<slug>.md`; the artifacts rules in `skills/SPEC.md` (8–9) and `writing-workflow-skills` name it as the home for every temp doc.
- **Cleanup** — removed the lingering `status: done` `TASK-chat-noise.md`; added `.thinkrail/` to the repo root `.gitignore`.

Decisions confirmed with the requester: host-seeds-per-workspace · temp-docs-only · keep the `.thinkrail/context` name.

## Verify

- `bun run check:deps` · `bun run lint` · `bun run typecheck` — green (also via pre-commit)
- `bun run test` — 139 server tests + all packages pass (incl. 2 new: scratch-dir seeding, `projectHasSpecs` task-spec exclusion)
- `bun run e2e` — 40/40 pass, incl. the workspace-creation and Welcome (`projectHasSpecs`) specs


OLavrik:
Nit: .thinkrail is hardcoded in two places (persistence.ts, workspaces.ts) with no shared constant. Suggest extracting it (e.g. to @thinkrail/shared) to simplify future renames. Non-blocking.

OLavrik (dismissed):

PR #61, opened 2026-07-11
feat: in-app provider authentication (OAuth, API key, JetBrains AI) in Settings

rsolmano:
## Summary

Turns model-provider authentication into a first-class, in-app experience. Previously the Welcome screen showed read-only status and told users to open a terminal (`pi` + `/login`, or `thinkrail jbcentral`). Now everything happens inside the app, in **Settings → Providers**:

- **Sign in with a subscription (OAuth)** — Anthropic (Claude Pro/Max), ChatGPT Plus/Pro, GitHub Copilot — driven through pi's `AuthStorage` in-process (open the URL or paste the code).
- **Add / remove an API key** for single-key providers.
- **JetBrains AI** — route Claude + GPT through your JetBrains subscription (the `central` proxy) with a guided state machine: not installed → not signed in → connect → connected.
- The Welcome screen keeps only a **slim warning** when no provider is connected, with a CTA that opens Settings → Providers.

Settings itself became a **two-pane shell** (section rail + content pane) so it can grow beyond Providers/GitHub.

## How it's structured (commit-by-commit)

1. `feat(contracts)` — the typed wire: login frames + `provider.login` channel, `provider.*` methods, status capability flags (`canOAuth`/`canApiKey`/`canLogout`), `jbcentralInstalled`, `PROTOCOL_VERSION` → 3.
2. `feat(server)` — the host engine: a new `auth/` submodule (provider status moved from `agent/`; a session-less `loginId`-keyed OAuth state machine over `AuthStorage.login`; API-key/logout; JetBrains AI connect/disconnect/login). The jbcentral protocol is consolidated into `@thinkrail/shared/jbcentral` and shared by `thinkrail jbcentral` and the in-app flow.
3. `feat(web)` — the UI: two-pane `SettingsDialog` + `ProvidersSettings` + `GithubSettings` + `JetBrainsAiCard`; the presentational session-less `LoginDialog`; the slim `ProviderWarningBanner`; store-held settings state.
4. `test(e2e)` — provider auth driven through Settings.
5–6. `test(e2e)` — **deterministic** coverage of the two hardest flows via test doubles: a stub `central` CLI (JetBrains AI connect round-trip) and a fake pi OAuth provider (full login: select → open-URL + paste → success → configured → sign-out, plus cancel).

## Notable

[...]

PR #83, opened 2026-07-16
fix(e2e): stop full-mode @agent specs cascading on shared host state

rsolmano:
## What

`bun run e2e:full` failed **19 `@agent` specs** locally. All three root causes are in the e2e harness — **no product code changed**.

## Root causes

**1. `jbcentral` spec poisoned the shared model registry (18 failures).**
On a machine authed through the jbcentral proxy, the seeded `models.json` *is* the auth (`anthropic`/`openai` `baseUrl`+`apiKey` wired to the proxy; `auth.json` is empty). The JetBrains AI connect/disconnect spec strips those fields via `unwireJbcentral` and leaves the host disconnected, and `resetState()` never restored `models.json` — so every `@agent` spec alphabetically after `jbcentral` found an empty registry and failed. (A dev authed via OAuth wouldn't hit this — hence machine-specific.)

**2. `error-turn` regressed from #61 (1 failure).**
#61 added host-side model validation (`resolveWireModel` throws on an unknown id at `session.create`). The spec rewired the model to a bogus tag expecting a provider 404 at prompt-time, but now `session.create` is rejected *before* a chat exists, so the failure surfaces as a **toast** ("Couldn't start the chat"), not an in-chat error turn.

**3. `setting-up-a-project` could delete the shared fixture repo (cascade).**
Once the registry was fixed and the agent actually ran, this spec drives a real agent with `bash` in a *worktree of the shared fixture repo*; a stray destructive command removed `/tmp/thinkrail-e2e/sample-project`, cascading into all later tests' `resetState`.

## Fixes (all in `e2e/`)

- **Re-seed `models.json` per test** — `globalSetup` snapshots the seeded `models.json` (`E2E_PI_MODELS_SEED`); `resetState()` restores it and drops the wire's `.bak`. The next page load's `provider.status` refreshes the registry, so `jbcentral` can't strip auth for the rest of the run.
- **`error-turn` asserts the kickoff toast** — `[data-testid="toast"][data-variant="error"]` carrying the rejected model id, no chat tab, no false "✓ Done".
- **Self-heal the fixture repo** — seeding moved to `e2e/fixtures/repo.ts` (`seedFixtureRepo` / `fixtureRepoHealthy`); `resetState()` re-seeds it when missing/damaged, bounding a flaky agent run's blast radius to that one spec.

## Result

Full suite: **19 systemic failures → 0**. Re-run was **71 passed / 1 failed**; the single miss (`visualize` comparison — the model didn't set the optional `recommended` flag that once) **passes in isolation** and is absorbed by CI's `retries: 1`. `check:deps` + `lint` + `typecheck` all pass.

## Test plan

- [x] `bun run e2e:full` — 19 → 0 systemic failures
- [x] `jbcentral` → `loader` → `multi-chat` in sequence: green (contamination fix)
- [x] `error-turn` in isolation: green
- [x] `visualize.live` in isolation: green (confirms remaining failure is `@agent` non-determinism)
- [x] `check:deps`, `lint`, `typecheck`


rsolmano:
Follow-up (commit e48969a): fixed a second, path-brittleness flake in `setting-up-a-project.live`.

Repro showed the agent **did** draft the full graph, but under a `specs/` dir (`specs/goal-and-requirements.md`, `specs/architecture.md`) instead of the worktree root — a legitimate agent choice the skill doesn't pin down. The test hard-asserted `existsSync(worktree/goal-and-requirements.md)` at the root, so it failed.

Fix: assert the graph root by frontmatter `type: goal-and-requirements` at **any depth** (the skill's real contract), not a hard-coded path. The Specs-viewer assertion already scans nested specs. Verified **3/3** across both root and `specs/` placements.

PR #86, opened 2026-07-17
feat(chat): ask_user_question survives host restarts — ack + terminate redesign

rsolmano:
## Problem

`ask_user_question` did not survive a host restart — and the failure was worse than losing the question:

- The tool's `execute` **blocked** on an in-memory promise (`pending`/`early` maps); pi persists the assistant message (with the `toolCall`) *before* the tool runs, so a restart left the transcript ending in an **unpaired tool call**.
- Every provider rejects a context with unpaired `tool_use` → **every later prompt 400'd; the chat was permanently bricked.**
- The rehydrated card still *looked* answerable, but the answer fell into a held-answers map nobody would ever consume ("Answer sent — continuing…" forever).
- Even a clean Ctrl+C hit this deterministically: shutdown called `process.exit(0)` on the same stack as dispose, so the cancelled tool result never landed on disk.

The two most popular open-source pi ask-extensions (`rpiv-ask-user-question`, `pi-ask-user`) have the same hole — they lean on the TUI's graceful-abort-before-exit, which our host violated and which never covers crashes (accepted in our architecture: in-process pi).

## Design: ack + terminate — asking ends the turn; the answer starts the next one

- `execute` validates, then **immediately** returns an ack (`details {kind:"ack"}`) with **`terminate: true`** — the turn ends at the tool batch with no extra LLM call. Nothing pends in memory; the transcript is always provider-valid; the session is genuinely **idle** while the user thinks (seconds or weeks, across any number of restarts).
- The reply (`session.answerQuestion`) is vetted against the transcript by the pure `assessAnswerability` (unknown call / already answered / not-awaiting / **superseded** → fail loud, never park) and injected via pi's public `AgentSession.sendCustomMessage({triggerTurn:true})` as an **`ask-user-answers` custom message** (same LLM envelope the blocking design used, `details {toolCallId, result}`). **Answering live and answering after a restart are the same code path.**
- A free-form user message instead of an answer **supersedes** the questionnaire — the card flips to a terminal record, and the model treats the message as the reply.
- Web: the card's lifecycle (*composing → awaiting → answered/declined/superseded/dead*) derives from the transcript via the new `askState` seam; legacy blocking-era transcripts still render correctly.

## Safety net (generic, not ask-specific)

- **Repair on re-open** (`sessionRepair.ts`): before a disk session seeds its context, every dangling tool call (crashed `bash`, `web_search`, old-format asks…) gets a synthetic result — pi's abort convention for generic tools, the canonical decline + re-ask hint for asks. Heals already-bricked transcripts.
- **Polite shutdown** (`boot.ts`): SIGINT/SIGTERM now abort streaming sessions and wait (bounded 2s) so pi persists their "Operation aborted" results before exit.

## Wire

- `TranscriptMessage` (pi-canonical + `custom` role) on `session.getMessages`; `AskUserQuestionAckDetails` / `AskUserAnswersDetails` / `ASK_USER_ANSWERS_CUSTOM_TYPE`; **protocol v7**.

## Specs & skills

Durable decisions promoted into `agent`, `contracts`, `host`, `chat`, `chat/tools`, `store` SPECs (including the rejected alternatives: the old blocking design, a suspended-session variant, bundling rpiv). `asking-user-questions` skill now teaches the turn-ending + supersede semantics.

## Verification

- `check:deps` + biome + typecheck + unit tests: green (new: repair against a real in-memory pi `SessionManager`, answerability verdicts, askState derivation, reducer/hydration indexing).
- e2e no-agent: **42/42**.
- e2e `@agent` (real provider): ask suite **9/9**, incl. a new supersede test and awaiting-card-survives-reopen (the exact hydration path a restart takes); full agent suite 34/35 pre-rebase — the one failure (`error-turn.live.spec.ts`) **also fails on main with this change stashed** (pre-existing, env-dependent model resolution at `session.create`; unrelated — worth a separate issue).

## Notes

- Possible upstream follow-up: `pi -c` after a SIGKILL mid-tool has the same unpaired-call brick — a load-time repair in pi would fix the whole ecosystem.
- Deferred by design: auto-decline timeout (pi-ask-user-style) — unnecessary once nothing blocks.

### DIFF
--- a/apps/web/src/chat/ChatActions.tsx
+++ b/apps/web/src/chat/ChatActions.tsx
@@ -3,15 +3,16 @@ import { createContext, useContext } from "react";
 
 /**
  * The interaction seam for chat tool renderers that need to talk BACK to the agent (e.g. the inline
- * `ask_user_question` card answering a blocked tool call). Presentational renderers stay store/transport
+ * `ask_user_question` card sending its reply). Presentational renderers stay store/transport
  * free — they read these callbacks from context instead. `ChatView` (the app-integration layer) provides
  * the value, wired to the transport + this tab's `sessionId`; when no provider is present (a renderer used
  * standalone, e.g. an extracted `packages/chat-ui`), the context is `null` and the card renders read-only.
  */
 export interface ChatActions {
 	/**
-	 * Answer a blocked `ask_user_question` tool call, correlated by its tool call id. Rejects when the host
-	 * refuses the reply (unknown session, malformed result) so the card can un-latch its "sent" state.
+	 * Answer an awaiting `ask_user_question` questionnaire, correlated by its tool call id — the host
+	 * injects the reply as the next turn's `ask-user-answers` message. Rejects when the host refuses it
+	 * (unknown session/call, already answered, superseded) so the card can un-latch its "sent" state.
 	 */
 	answerQuestion: (toolCallId: string, result: AskUserQuestionResult) => Promise<void>;
 }
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -9,6 +9,7 @@ import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Virtuoso, type VirtuosoHandle } from "react-virtuoso";
 import { EMPTY_RUNTIME, useAppStore } from "@/store";
 import { errorText, getTransport } from "@/transport";
+import { AskStatesContext, deriveAskStates } from "./askState";
 import { type ChatActions, ChatActionsContext } from "./ChatActions";
 import { ChatHeader } from "./ChatHeader";
 import { Composer, type MentionCandidate, type SubmitBehavior } from "./Composer";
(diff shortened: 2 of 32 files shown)


rsolmano:
**Added true host-restart e2e coverage** (`e2e/ask-restart.live.spec.ts`, @agent): a self-contained spec with its own host lifecycle (private port 24254 + isolated state dir + pi agent dir cloned from the suite seed) — the one scenario the shared-host suite structurally cannot cover.

Flow it pins: questionnaire open → **kill -9 the host** → fresh host on the same on-disk state → chat reopens from history (the hydration path) → the awaiting card is **still answerable** → the answer starts a new turn and the agent replies with it in context.

Full verification re-run on the branch:
- no-agent suite: **42/42**
- `@agent` suite: **37/37** — including this new restart spec, all 9 ask tests, and `error-turn.live.spec.ts`, which passed this time (confirming the earlier failure noted in the PR description was flake, not a regression).

danyaberezun (approved):

PR #88, opened 2026-07-18
feat: add project removal without touching the source repo (#53)

### LINKED ISSUE #53: Add project removal (remove all workspaces, keep the source directory untouched)
## Summary

ThinkRail lets you open a **project** (a git repository) and create **workspaces** inside it — each a `git worktree` on its own branch under `~/.thinkrail/worktrees/…`, never inside your source repo. Today you can remove individual workspaces, but there is **no way to remove a whole project** — once opened it stays in the left rail forever.

Add a **Remove project** action that:

- removes **all of the project's workspaces** — their worktrees on disk, their chat sessions, and their terminals; and
- leaves the **source directory completely untouched** — no changes to the repo's working tree, files, or branches.

## Proposed approach

Mirror the existing single-workspace removal (`workspace.remove` in `packages/server/src/host/handlers.ts`): the same non-blocking archive, applied to every workspace in the project, then drop the project record.

- **Wire:** rename the dead `project.close` method → `project.remove` (symmetry with `workspace.remove`; `project.close` has no caller today).
- **Host:** a `project.remove` handler does the fast part synchronously (per workspace: forget record → evict spec index → close terminals; then forget the project), acks, and runs the slow teardown in the background (per workspace: dispose sessions + purge transcripts → `git worktree remove`). The composition lives in the **host** because the `projects` module must not import `workspaces`.
- **Keep git branches.** Worktree dirs are reclaimed, chats/terminals purged, records dropped — but the branches in the source repo stay untouched.
- **UI:** kebab (…) menu on the project row → destructive `ConfirmDialog`; the store removes the project **optimistically** (drop it + its workspaces' tabs/sessions/terminals, return to Welcome), reconciling a failure by re-listing.
- ⚠️ **Gotcha:** `reclaimWorktree` finds the repo path via the project record — which this teardown drops *first*. Give it an optional `repoPath` parameter, captured at the start of the handler.

## Acceptance criteria

- [ ] A project can be removed from the left rail via a kebab (…) menu + a destructive confirmation.
- [ ] Removing a project removes **all** its workspaces: worktrees reclaimed, chat sessions disposed + on-disk transcripts purged, terminals closed.
- [ ] The source repository is **not modified** (working tree clean, files intact, branches kept).
- [ ] Removing the currently-selected/active project returns the UI to the Welcome screen.
- [ ] Removal is optimistic and non-blocking (the row vanishes immediately; the git reclaim runs in the background), reconciling a failure by re-listing.
- [ ] `lint`, `typecheck`, `test`, and `e2e` all pass; a new e2e case covers the flow.

dheerajp45:
## Summary

Adds **project removal** so a project can leave the left rail without deleting the user's source repository.

Today users can remove individual workspaces, but an opened project stays forever. This PR adds a **Remove project** flow that:

- **Wire:** renames unused `project.close` → `project.remove` (protocol v8), symmetric with `workspace.remove`
- **Host:** archives every workspace in the project (fast: forget record, evict spec index, stop watch, close terminals; then drop the project record; slow/background: dispose sessions, purge transcripts, `git worktree remove`), composed in the host so `projects` never imports `workspaces`
- **Gotcha:** captures `repoPath` before dropping the project and passes it into `reclaimWorktree` so background reclaim still works
- **Keeps git branches** and leaves the source working tree/files untouched
- **UI:** kebab (`…`) on the project row → destructive `ConfirmDialog`
- **Store:** optimistic removal (project + workspace tabs/sessions/terminals, back to Welcome) with re-list on failure
- **Tests:** unit coverage for store/reclaim path + e2e for the full remove flow

## Related issues

Closes #53

## Checklist

- [ ] Fast gates pass: `bun run lint`, `bun run typecheck`, `bun run test`


dheerajp45:
Ready for review.

Addressed Copilot feedback (`listWorkspaceRecords` for teardown, e2e poll timeout, SPEC/spacing cleanups) and rebased onto `main` (protocol bump is **v8** since settings already took v7).

CI is waiting on maintainer workflow approval — happy to follow up on any review comments. Thanks!

Closes #53

rsolmano:
@dheerajp45 Sorry for the long waiting review time. Please resolve merge conflicts and I'll jump to review.

dheerajp45:
@rsolmano Conflicts resolved (protocol v9). Ready for review — thanks!

rsolmano (changes requested):
**Overall this is clean, well-scoped work.** The design mirrors the issue's proposed approach closely, the `projects` ⊄ `workspaces` boundary is respected (the archive composition lives in the host), the `repoPath`-before-`closeProject` gotcha is handled, and the e2e convincingly proves the source repo is left untouched. Nice.

One thing to cover before merge: **multi-client (multi-tab) convergence for the project itself.**

Scenario — workspace *W* of project *P* is open and active in **Tab A**; you remove *P* from **Tab B**:

- ✅ The **workspace** side converges correctly: `forgetWorkspace` fans out `workspace.removed` to every client, so Tab A's row disappears, its tabs/terminals/chat clear, and it drops to Welcome.
- ❌ The **project** side does not: there's no `project.removed` push, so **Tab A keeps a stale, empty "ghost" project row** until it reloads/reconnects. Interacting with the ghost is degraded (it lists 0 workspaces; creating a workspace throws `Unknown project`).

Could you make project removal converge across clients the same way workspace removal already does? The workspace path is the reference to mirror:

- the `workspace.removed` channel in `WS_CHANNELS` (`packages/contracts/src/wsProtocol.ts`),
- the `workspaces` module's injected publisher (`setWorkspacePublisher`) that emits the lifecycle event, and the host mapping `kind → channel`,
- the store reacting to the push in `applyWorkspaceRemoved` (`apps/web/src/store/appStore.ts`) — every client reacts identically, rather than relying only on the initiator's optimistic `removeProject`.

Concretely: add a `project.removed` broadcast (payload `{ id }`), publish it from the `project.remove` handler after `closeProject`, and have the store clear the row (+ its workspaces/tabs/terminals, returning to Welcome if selected/active) on that push for every client. Keeping the initiator's optimism as a local fast-path is fine (the push handler just needs to be idempotent). Please update the relevant `SPEC.md`s to reflect the new channel.

Also minor, while you're here: `apps/web/src/panels/SPEC.md` picked up a small merge artifact from this change — a duplicated sentence (`surfaces an error toast, leaving the row in place).`) and an unbalanced trailing paren around the new "Project removal" paragraph. Since specs are the source of truth here, worth tidying.

Finally, could you attach a **short screen recording** of the two-tab flow confirming it now converges — remove *P* in one tab and show the project row **and** its workspaces disappearing in the other tab, with no reload?

Thanks!

dheerajp45:
Thanks — that makes sense. I'll add a `project.removed` channel mirroring `workspace.removed` (broadcast after `closeProject`, store `applyProjectRemoved` for every client, keep initiator optimism idempotent), tidy the panels SPEC merge artifact, and attach a two-tab screen recording.

dheerajp45:
Added `project.removed`


 (and `project.opened` for symmetry with workspace create) so other tabs drop/add the project row without a reload.
Protocol: v9 `project.remove`, v10 `project.removed`, v11 `project.opened` (v8 stays `ask_user_question`).
Screen recording attached: two tabs open → remove project in one → other tab updates with no refresh.

https://github.com/user-attachments/assets/5b964c07-e64c-4459-9e7d-194170fb6771

rsolmano:
Thanks for the update—this is a substantial improvement. I verified that the happy-path project removal now works: project/workspace records and worktrees are removed, live terminals are terminated, source files and working tree remain untouched, branches are retained, and `project.opened` / `project.removed` correctly converge across browser tabs.

I also reran the dependency, lint, typecheck, unit, and browser e2e suites successfully.

I can’t approve yet because several destructive failure and concurrency cases still need to be addressed:

### 1. `workspace.remove` followed by `project.remove` can leak a worktree

`workspace.remove` starts `archiveTeardown(ws)` without carrying the repository path. If `project.remove` drops the project while session cleanup is awaiting, `reclaimWorktree` can no longer resolve the source repository and silently leaves the worktree behind.

**Suggested fix:** capture the repository path before starting the detached teardown and pass it explicitly:

```ts
const ws = forgetWorkspace(id);
const repoPath = ws
  ? listProjects().find((project) => project.id === ws.projectId)?.path
  : undefined;

evictSpecIndex(id);
stopWatch(id);
closeWorkspaceTerminals(id);

if (ws) void archiveTeardown(ws, repoPath);
```

Please also add a regression test that sends `workspace.remove` immediately followed by `project.remove` and waits until `git worktree list` no longer contains the removed worktree.

### 2. In-flight `workspace.create` can finish after project removal

`createWorkspace` loads `workspaces.json` before its asynchronous fallback fetch. While that fetch is awaiting, `project.remove` can delete the project and its workspaces. Creation can then resume with stale project and workspace snapshots, create a worktree, and persist a workspace for a nonexistent project. The stale workspace snapshot can also resurrect records removed during the await.

**Suggested fix:**

1. Introduce a host-level project-removal guard/tombstone—or a per-project lifecycle mutex—shared by `workspace.create`, `workspace.remove`, and `project.remove`.
2. After every asynchronous gap in `createWorkspace`, re-resolve the project before performing mutations.
3. Load the current workspace registry immediately before saving rather than retaining the pre-await snapshot.

For example:

```ts
if (baseNeedsFetch) {
  await gitAsync(project.path, ["fetch", "origin", "--", branch]);
}

const currentProject = getProjects().find((p) => p.id === projectId);
if (!currentProject || removingProjects.has(projectId)) {
  throw new Error(`Project was removed during workspace creation: ${projectId}`);
}

// Do not reuse the registry loaded before the await.
const currentWorkspaces = loadWorkspaces();
// Create and persist using currentProject/currentWorkspaces.
```

If worktree creation succeeds but a subsequent validation or persistence operation fails, the newly created worktree should also be reclaimed before returning the error.

### 3. Persistence failure can leave a partially removed project

`project.remove` forgets workspace records one at a time and only afterward removes the project and schedules background teardown. If `closeProject` or another fast-path operation throws, previously forgotten workspaces are never restored and their teardown jobs are never scheduled.

**Suggested fix:** introduce a bulk project-archive operation rather than composing repeated independent persistence writes:

- Load the project and all child workspaces once.
- Persist the new project/workspace registry state once per file, preferably through atomic temp-file-and-rename writes.
- Return the captured workspace records and repository path to the host.
- Emit lifecycle events only after the registry transition succeeds.
- Ensure every captured/forgotten workspace is scheduled for cleanup even if a later non-persistence teardown step fails.

Ideally, use a persisted removal intent/journal so a process crash or failure between the two registry files can be recovered on startup. At minimum, add explicit rollback and a `finally` path so a workspace can never be forgotten without either being restored or having its teardown scheduled. This would also eliminate the current repeated `workspaces.json` rewrites during bulk removal.

### 4. Failed optimistic removal irreversibly destroys client context

If the `project.remove` request fails before reaching the host, reconciliation restores the project and workspace rows, but the active workspace, editor tabs, terminal state, and live PTY have already been discarded.

**Suggested fix:** separate optimistic presentation from authoritative resource disposal:

- Track `pendingProjectRemovals` separately.
- Hide or mark the row as removing immediately, without calling `clearWorkspaceTabs` or deleting terminal/session state.
- On request failure, remove the pending marker and reveal the unchanged project state.
- Only call `applyProjectRemoved`—including tab, session, and terminal cleanup—after the authoritative `project.removed` event or successful acknowledgement.

If retaining the optimistic disappearance is too complex, the safer alternative is to follow the existing `workspace.remove` design: send the request and let the lifecycle event perform all local cleanup.

A store snapshot alone is not sufficient if optimistic cleanup has already closed the host PTY; that process cannot be restored from browser state. Please add browser coverage that injects a rejected `project.remove` request and verifies that the project, active workspace, editor tabs, terminal tab, and live terminal process all remain intact.

Once these lifecycle and rollback cases are addressed and covered by regression tests, I’d be happy to approve.

dheerajp45:
Thanks for the thorough review — it was really helpful.

I'm going to pause work on this PR for now rather than leave the concurrency fixes half-finished. I'm happy for someone else to pick it up from here, or I may come back to it later once I'm more familiar with this area.

Thanks again for the detailed verification and the clear guidance on how to fix the issues.

Issue #98, opened 2026-07-21
Background model-catalog refresh on model.list (pi ≥0.80.8 live catalogs)


rsolmano:
## Context

pi 0.80.8 added live dynamic catalogs: `ModelRuntime.refresh()` fetches provider catalog overlays (network-throttled to ~4h internally; `pi update --models` forces it), and pi's own `/model` renders the current snapshot immediately while refreshing in the background. Our host composes catalogs at runtime creation and on `provider.status` reads (`reloadConfig`), but `model.list` never triggers a catalog refresh — newly-shipped catalog models don't appear in the picker until a host restart.

## Proposal

Mirror pi's `/model` in `packages/server`: `model.list` returns the current availability immediately and fires a **detached** `runtime.refresh()`; a later read picks up new models.

## Acceptance criteria

- [ ] `model.list` latency is unchanged (never awaits the network refresh).
- [ ] The background refresh respects pi's internal throttle; failures are logged and never fail the picker request.
- [ ] A new catalog model appears on a later `model.list` without restarting the host.
- [ ] Unit test pins the fire-and-forget (list result unaffected by a hanging refresh).
- [ ] `packages/server/src/agent/SPEC.md` notes the refresh trigger.

Issue #97, opened 2026-07-21
Drive API-key provider setup through the interactive login channel (pi ≥0.80.8 provider-owned login)


rsolmano:
## Context

pi 0.80.8 made provider auth metadata public (`Provider.auth: {apiKey?, oauth?}`) and drives API-key setup through the same `AuthInteraction` prompt flow as OAuth (`runtime.login(id, "api_key", interaction)`) — including providers that need more than one string (and `amazon-bedrock`, which now accepts a single API key). Our strip only offers an inline single-string field, gated by hand-maintained exclusion sets (`MULTI_FIELD_PROVIDERS` / `OAUTH_ONLY_PROVIDERS` in `packages/server/src/auth/providerStatus.ts`). The WS login bridge (`provider.loginStart` → `LoginFrame` dialog) already renders `select`/`prompt` frames generically, so it can carry api-key flows as-is.

## Proposal

Offer **Sign in with API key** through the existing login dialog for every provider whose `Provider.auth.apiKey.login` exists, then delete the exclusion sets — capability flags derive from `Provider.auth` alone.

## Acceptance criteria

- [ ] A provider with `Provider.auth.apiKey.login` can be configured via the login dialog's prompt flow — including `amazon-bedrock` (single key) and `azure-openai-responses` (multi-prompt).
- [ ] `MULTI_FIELD_PROVIDERS` / `OAUTH_ONLY_PROVIDERS` are removed; `canApiKey`-style flags come from `Provider.auth` metadata only.
- [ ] `openai-codex` (no apiKey auth) still offers no API-key entry.
- [ ] Keys persist to auth.json (survive a host restart) — via `runtime.login`, never `setRuntimeApiKey`.
- [ ] e2e covers one API-key login through the dialog (fake provider is fine).
- [ ] `packages/server/src/auth/SPEC.md` updated (it documents the sets and names this issue).

PR #99, opened 2026-07-21
feat(server): migrate to pi 0.80.10 ModelRuntime

rsolmano:
## What

Bumps `@earendil-works/pi-{coding-agent,ai,agent-core}` **0.80.6 → 0.80.10** and migrates `packages/server` fully onto the new `ModelRuntime` facade. pi 0.80.8 replaced the SDK's auth/model surface (`CreateAgentSessionOptions.authStorage`/`modelRegistry` → async `modelRuntime`; `AuthStorage` no longer exported; `getApiKeyAndHeaders` → `getAuth`; refresh went async), so this is a real migration, not a version bump.

## How

| Area | Migration |
|---|---|
| `agent/piRuntime` | One shared `ModelRuntime` — lazily-memoized async create (preserves the `PI_CODING_AGENT_DIR`-at-first-use property the e2e harnesses rely on); a failed create clears the memo so the next call retries |
| `agent/agentSessionManager` | `createAgentSession({modelRuntime})`; model reads via async `getAvailable()` |
| `agent/oneshot` | `runtime.completeSimple()` — deletes the spec-flagged `pi-ai/compat` `complete()` surface and manual auth resolution (auth incl. OAuth refresh now resolves inside pi's request path) |
| `auth/providerStatus` | `reloadConfig()` revalidation — pi's file credential store reads auth.json live under a lock, so the old `reload()` dance dies; OAuth rows derive from the now-public `Provider.auth.oauth` |
| `auth/providerLogin` | Old `OAuthLoginCallbacks` → `AuthInteraction` (`notify`/`prompt`, incl. pi cancelling the loser of its browser-vs-paste race via `prompt.signal`, identity-guarded parking). **API keys persist via `login(id, \"api_key\")`** — `setRuntimeApiKey` turned out to be a non-persistent overlay that would have silently dropped keys on restart |
| `dev.ts` | Fake e2e OAuth provider via `runtime.registerProvider` — extension registrations survive refresh now, so the `ModelRegistry.prototype.refresh` monkey-patch is deleted |
| Misc | `StringEnum` moved to the pi-ai root (×3); `provider.setApiKey`/`provider.logout` handlers async; SPECs (agent + auth) updated with the code |

## Regression found & fixed during e2e

0.80.10's catalog gives `openai-codex` model rows, so the old model-rows-based `canApiKey` wrongly offered a key field to an OAuth-only provider. `ProviderStatusSources` gained an `apiKeyLogin(id)` source (pi's public `Provider.auth.apiKey.login` truth), AND-ed with the kept conservative exclusion sets — exactly the drift the auth SPEC's note warned to re-check on pi bumps. The sets' removal is tracked in #97.

## Follow-ups (filed, not in this PR)

- #97 — drive API-key setup through the interactive login channel (deletes the exclusion sets; covers Bedrock's new single-key path)
- #98 — background model-catalog refresh on `model.list` (pi 0.80.8 live catalogs)

Free wins active with the bump: prompt-cache no longer invalidates daily (pi removed the date from its system prompt), Grok 4.5 + xAI device-code OAuth (works through the existing login dialog), Kimi K3, Fable 5 `xhigh`/`max` thinking levels.

## Verification

- `bun run check:deps` ✓ · `bun run lint` ✓ · `bun run typecheck` ✓
- `bun run test` — **168/168** (providerLogin rewritten for `AuthInteraction`; oneshot + session-manager tests now drive a **real** `ModelRuntime` with in-memory credentials + faux providers through pi's actual auth/dispatch path; new `canApiKey` pin for the openai-codex case)
- `bun run e2e:full` — **80/81 + 1 retry-pass** (the one failure was an `@agent` LLM flake: the model ignored \"mark exactly one option as recommended\"; the spec passed standalone in 7.6s). Includes the in-app fake-OAuth login flow against the reworked `dev.ts`.
- Suppression audit clean (no new `biome-ignore`/`as any`/`@ts-*`); `spec_validate` ✓

No wire/protocol changes; no UI-visible changes.

### DIFF
--- a/packages/server/src/agent/oneshot.ts
+++ b/packages/server/src/agent/oneshot.ts
@@ -1,15 +1,12 @@
 /**
  * One-shot LLM completions — the primitive behind the ad-hoc "assist" tasks (workspace naming, PR
- * drafting). It runs a **single** `complete()` request on a cheap, already-authenticated model: no
- * `AgentSession`, no tools, no extensions, nothing written to disk. This mirrors how pi itself runs its
- * compaction summaries — pick a model from the shared runtime's authenticated set, resolve its auth
- * (OAuth refresh included), and dispatch one request.
- *
- * `complete` is imported from `@earendil-works/pi-ai/compat`, the api-dispatch entry pi's own model
- * plumbing uses. It's a compat surface pi will fold into its ModelManager migration — re-verify this
- * import on every pi bump (the `ModelRegistry` half we rely on for model + auth resolution is stable).
+ * drafting). It runs a **single** completion on a cheap, already-authenticated model: no
+ * `AgentSession`, no tools, no extensions, nothing written to disk. Dispatch goes through the shared
+ * `ModelRuntime.completeSimple()` — pi's canonical provider-agnostic request path (it also serves
+ * providers that only implement `streamSimple`, e.g. extension-registered ones), which resolves the
+ * model's auth itself (OAuth refresh included), so there is no separate auth-resolution step here.
  */
-import { type Api, type Context, complete, type Model } from "@earendil-works/pi-ai/compat";
+import type { Api, Context, Model } from "@earendil-works/pi-ai";
 import { getPiRuntime } from "./piRuntime";
 
 /** Which model a one-shot task reaches for. `cheap` = small/fast; `default` = first authenticated. */
@@ -58,8 +55,8 @@ const CHEAP_MODELS: ReadonlyArray<readonly [provider: string, idPrefix: string]>
  * For `default`: the first authenticated model. `null` when nothing is authenticated — the caller
  * degrades gracefully rather than erroring.
  */
-export function pickModel(tier: ModelTier = "cheap"): Model<Api> | null {
-	const available = getPiRuntime().modelRegistry.getAvailable();
+export async function pickModel(tier: ModelTier = "cheap"): Promise<Model<Api> | null> {
+	const available = await (await getPiRuntime()).getAvailable();
 	if (available.length === 0) return null;
 	if (tier === "default") return available[0] ?? null;
 	for (const [provider, prefix] of CHEAP_MODELS) {
@@ -80,21 +77,15 @@ export function pickModel(tier: ModelTier = "cheap"): Model<Api> | null {
  * assist tasks) wrap this and fall back.
  */
 export async function completeOnce(req: OneShotRequest): Promise<OneShotResult> {
-	const { modelRegistry } = getPiRuntime();
-	const model = pickModel(req.tier);
+	const runtime = await getPiRuntime();
+	const model = await pickModel(req.tier);
 	if (!model) throw new Error("no-model");
-
-	const auth = await modelRegistry.getApiKeyAndHeaders(model);
-	if (!auth.ok) throw new Error(auth.error);
 
 	const context: Context = {
 		...(req.system ? { systemPrompt: req.system } : {}),
 		messages: [{ role: "user", content: req.prompt, timestamp: Date.now() }],
 	};
-	const message = await complete(model, context, {
-		...(auth.apiKey ? { apiKey: auth.apiKey } : {}),
-		...(auth.headers ? { headers: auth.headers } : {}),
-		...(auth.env ? { env: auth.env } : {}),
+	const message = await runtime.completeSimple(model, context, {
 		maxTokens: req.maxTokens ?? 256,
 		temperature: req.temperature ?? 0.2,
 		...(req.signal ? { signal: req.signal } : {}),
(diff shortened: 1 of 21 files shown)


packages/server/src/auth/providerLogin.ts
```diff
@@ -37,88 +38,105 @@ function terminate(loginId: string): Pending | undefined {
...
+			kind: "select",
+			message: prompt.message,
+			options: prompt.options.map((o) => ({ id: o.id, label: o.label })),
+		};
+	}
+	// `text`, `secret`, and `manual_code` all collect one string (the paste-code race included).
+	return {
+		kind: "prompt",
+		message: prompt.message,
+		...(prompt.placeholder ? { placeholder: prompt.placeholder } : {}),
```
danyaberezun, line 56:
```suggestion
		...(prompt.placeholder ? { placeholder: prompt.placeholder } : {}),
        ...(prompt.type === "text" ? { allowEmpty: true } : {}),
```

otherwise breaks default option for copilot (github)

packages/server/src/auth/providerStatus.ts
```diff
@@ -156,32 +164,39 @@ export function buildProviderReport(sources: ProviderStatusSources): ProviderSta
...
+ * restarting the host. (Accepted micro-risk: refreshing the shared runtime concurrent with a streaming
+ * session — the same thing pi's TUI does on `/login`.)
  */
-export function getProviderStatus(): ProviderStatusReport {
-	const { authStorage, modelRegistry } = getPiRuntime();
-	authStorage.reload();
-	modelRegistry.refresh();
+export async function getProviderStatus(): Promise<ProviderStatusReport> {
+	const runtime = await getPiRuntime();
+	await runtime.reloadConfig();
```
danyaberezun: `reloadConfig()` doesn't re-read `auth.json` --- this comment (and `auth/SPEC.md`) claim it does; `reloadConfig()` reloads `models.json` only.

packages/server/src/agent/oneshot.ts
```diff
@@ -58,8 +55,8 @@ const CHEAP_MODELS: ReadonlyArray<readonly [provider: string, idPrefix: string]>
  * For `default`: the first authenticated model. `null` when nothing is authenticated — the caller
  * degrades gracefully rather than erroring.
  */
-export function pickModel(tier: ModelTier = "cheap"): Model<Api> | null {
-	const available = getPiRuntime().modelRegistry.getAvailable();
+export async function pickModel(tier: ModelTier = "cheap"): Promise<Model<Api> | null> {
+	const available = await (await getPiRuntime()).getAvailable();
 	if (available.length === 0) return null;
 	if (tier === "default") return available[0] ?? null;
 	for (const [provider, prefix] of CHEAP_MODELS) {
```
danyaberezun, line 60: Stale xAI `cheap`- tier hints -dead after this
, so please update to a current small xAI id or drop the two lines.

danyaberezun (changes requested):

danyaberezun:
---
Doc drift: stale AuthStorage/ModelRegistry/compat references the migration missed.

The code and the agent/auth SPECs were migrated, but four references in untouched files still describe the removed two-object (AuthStorage + ModelRegistry) world. Per "the spec leads the code," please update:

- packages/server/SPEC.md:73 — "the shared AuthStorage + ModelRegistry" → the shared ModelRuntime.
- packages/shared/SPEC.md:60 — "adds modelRegistry.refresh()" → reloadConfig() (what auth/jbcentral.ts now calls).
- packages/spec-graph/tools/SPEC.md:25 — says StringEnum comes from @earendil-works/pi-ai/compat; this PR moved it to the @earendil-works/pi-ai root (see create.ts/graph.ts).
- packages/server/src/host/server.ts:171 — comment "the session-less authStorage.login bridge" → ModelRuntime.login.

---

packages/server/src/auth/providerLogin.ts
rsolmano: Applied in fa403ee — `text`-type prompts now carry `allowEmpty: true` (verified against pi-ai 0.80.10: the Copilot GHE prompt is `type: "text"` with blank = github.com, and our LoginDialog only permits an empty submit when the frame carries the flag). `secret`/`manual_code` stay strict since an empty key/auth-code is never meaningful. Unit test updated to pin the flag on the Copilot-path frame.

packages/server/src/auth/providerStatus.ts
rsolmano: You're right — `reloadConfig()` only reloads models.json. The reason external logins still appear is one hop further: `reloadConfig()` ends in `refresh()`, whose availability refresh re-runs the per-provider auth checks + `credentials.list()` against pi's credential store, and that file store re-reads auth.json (under a lock) on every access. Reworded the comment and `auth/SPEC.md` in fa403ee to attribute it precisely ('does NOT touch auth.json itself; its internal availability refresh re-runs the auth checks…').

packages/server/src/agent/oneshot.ts
rsolmano: Fixed in fa403ee — probed the 0.80.10 xAI catalog (`grok-4.3`, `grok-4.5`, `grok-build-0.1`): both hints were dead, replaced with the current small tier `xai/grok-build` (prefix-matched, so it survives 0.1→0.x). Also swept the rest of the allowlist the same way and dropped `anthropic/claude-3-5-haiku` (gone from the catalog too; the `claude-haiku` prefix already covers haiku-4-5).

danyaberezun (approved):

Issue #19, opened 2026-07-06
Diff viewer: split viewer for diff


rsolmano:
We need to have some view to render changes

PR #100, opened 2026-07-22
Diff viewer: split viewer for diff  #19

OLavrik:
Screenshot 2026-07-22 at 12 46 36


packages/server/src/git/git.ts
```diff
@@ -95,38 +136,52 @@ export function gitStatus(workspaceId: string): GitStatus {
...
 
 	const untracked = git(ws.worktreePath, ["ls-files", "--others", "--exclude-standard"]);
 	if (untracked.ok && untracked.out) {
 		for (const path of untracked.out.split("\n")) {
-			if (path) changes.push({ path, status: "untracked" });
+			if (!path) continue;
+			// Untracked files never appear in `git diff` — count their whole content as added.
+			let added = 0;
+			try {
+				added = lineCount(readFileSync(resolve(ws.worktreePath, path), "utf8"));
```
rsolmano: **Major — untracked binaries get garbage counts, and this read is unbounded.**

Tracked binaries are correctly dropped (numstat reports `-`/`-` and you skip them), but every *untracked* file — binary or not — is read fully as UTF-8 and line-counted. Verified live: an untracked PNG shows a `+3` badge in both Changes views.

Two problems:
1. **Wrong data** for binaries (mojibake "lines" counted as additions) — inconsistent with the tracked side within the same list.
2. **Perf/memory**: this runs on *every* `git.status` (i.e. every fs tick while the Changes tab is open). A large untracked artifact (build output, archive) is re-read into memory each time, synchronously on the event loop.

Suggest a size cap (`statSync` first) + a cheap binary sniff (NUL byte in the first few KB → omit counts, matching the tracked-binary behavior).

packages/server/src/git/git.ts
```diff
@@ -95,38 +136,52 @@ export function gitStatus(workspaceId: string): GitStatus {
...
-		// `git diff --no-index` exits non-zero when files differ; the patch is still on stdout.
-		return {
-			diff: git(ws.worktreePath, ["diff", "--no-color", "--no-index", "/dev/null", path]).out,
-		};
+	const abs = resolve(ws.worktreePath, path);
+	const rel = relative(ws.worktreePath, abs);
+	if (rel.startsWith("..") || isAbsolute(rel)) throw new Error("Path escapes the worktree");
+
+	const base = git(ws.worktreePath, ["show", `${ws.baseBranch}:${path}`], { raw: true });
+	const original = base.ok ? base.out : "";
```
rsolmano: **Minor — any `git show` failure reads as "file absent in base".**

`base.ok ? base.out : ""` conflates the intended case (path not in the base branch → add-style diff) with *every other* failure: index lock contention, a deleted/invalid base ref, repo corruption. All of them silently render a whole-file-added diff, which actively misleads.

Worth distinguishing (e.g. treat only the "does not exist in" / "invalid object name" stderr as absence; `console.warn` anything else), so a broken read stays observable. Related nit: for renames the `+/−` badge shows the true delta while this pane shows whole-file-added — documented, but visibly self-contradictory; a sentence in the pane header ("renamed from …") would reconcile it cheaply later.

apps/web/src/panels/ChangesPanel.tsx
```diff
@@ -87,18 +48,48 @@ export function ChangesPanel({ workspaceId }: { workspaceId: string }) {
...
+	// status list is loaded — the diff opens only on the user's explicit click. Match by suffix so an
+	// absolute pi path still resolves to the relative entry.
 	useEffect(() => {
 		if (!status || changesRequest?.workspaceId !== workspaceId) return;
 		const want = changesRequest.path;
 		// Anchor the suffix at a path separator so an absolute pi path resolves to its relative entry
 		// without `a-foo.ts` spuriously matching the entry `foo.ts`.
 		const match = status.changes.find((c) => c.path === want || want.endsWith(`/${c.path}`));
-		void selectFile(match ? match.path : want);
+		setHighlighted(match ? match.path : want);
```
rsolmano, line 86: **Blocker (test not migrated for this contract change).**

Changing the chip deep-link from "select + show diff" to highlight-only is a defensible, spec-recorded call — but `e2e/turn-divider.live.spec.ts` still pins the *old* contract and the deleted `diff-viewer` testid, so the `@agent` suite is deterministically red (verified: passes on main, fails on this branch merged with main). Please update that spec to assert the new behavior (Changes tab active + the `change-item` row highlighted) in the same PR that changes the contract.

apps/web/src/panels/ChangesPanel.tsx
```diff
@@ -87,18 +48,48 @@ export function ChangesPanel({ workspaceId }: { workspaceId: string }) {
...
 	}, [changesRequest, status, workspaceId]);
 
+	// A row is selected when its diff tab is the active center tab, or it's the deep-link highlight.
+	const isActive = (path: string) =>
```
rsolmano, line 100: **Minor UX bug — two rows can render "selected" at once.**

`highlighted` is never cleared when the active center tab changes, and `isActive` ORs the two signals. Repro (verified live): click row A (opens tab A), click row B (opens tab B), switch the center tab back to A → both A (active tab) and B (stale highlight) show the selected background.

Suggest clearing `highlighted` whenever `activeTabId` changes to a diff tab of this workspace, or making the active-tab match the sole selection source once any diff tab is active (keeping `highlighted` only as the deep-link affordance while no tab is open).

apps/web/src/panels/DiffPane.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * Live: same contract as `FilePane` — when the workspace's fs tick moves past the tick this tab's
+ * contents were loaded at, both sides are re-read (`git.diffFile`) and replaced. Only the active tab
+ * mounts, so background tabs catch up on activation. A file that left the change set keeps its last
+ * contents (the Changes list is where the disappearance shows); a failed re-read just advances the tick.
+ */
+export function DiffPane({ tab }: { tab: DiffTab }) {
+	const setDiffTabView = useAppStore((s) => s.setDiffTabView);
+	const change = useAppStore((s) => s.fsChangesByWorkspace[tab.workspaceId]);
+
+	useEffect(() => {
```
rsolmano: **Refactor — this effect is a wholesale copy of `FilePane`'s.**

The staleness check, the skippable-single-batch optimization, the cancel flag, the advance-tick-on-failure — all duplicated line-for-line from `FilePane`, with only the read method and store action differing. The repo bar (AGENTS.md: "no duplicated derivations") says centralize: a small `useLiveTabContent(tab, read, apply)` hook in `panels/` would keep the two live-refresh contracts from drifting apart — especially since this PR just fixed the SPEC to describe them as one contract.

apps/web/src/panels/MonacoDiff.tsx
```diff
@@ -0,0 +1,55 @@
...
+			language={languageForPath(path)}
+			originalModelPath={`diff-original://${path}`}
+			modifiedModelPath={`diff-modified://${path}`}
+			theme={THEME}
+			beforeMount={beforeMount}
+			onMount={onMount}
+			loading={
+				<div className="flex h-full items-center justify-center text-hint">Loading diff…</div>
+			}
+			options={{ ...sharedEditorOptions(), renderSideBySide: view === "split" }}
```
rsolmano: **Minor UX — "Split" silently renders inline at narrow widths.**

Monaco's `useInlineViewWhenSpaceIsLimited` defaults to `true`, so below ~900px editor width the diff renders inline while the header toggle still shows **Split** active — the toggle looks like a no-op (verified at a 1280px viewport, where the center pane is ~660px; at 1920px split renders correctly). Either pass `useInlineViewWhenSpaceIsLimited: false` (respect the explicit per-tab choice — it's what the toggle promises) or reflect the effective mode in the toggle state.

apps/web/src/store/SPEC.md
```diff
@@ -95,12 +95,17 @@ editor tabs + terminals (switching workspaces swaps both), and a **per-session c
   change (the store holds only the signal, never fetches; `applyWorkspaceRemoved` drops a removed
   workspace's entry); and **`updateFileTabContent(id, content,
   tick)`** — a `FileTab` carries the `tick` its content was loaded at, so `FilePane` detects staleness
-  (`workspaceTick > tab.loadedTick`) across tab switches. The transient **`changesRequest`** +
+  (`workspaceTick > tab.loadedTick`) across tab switches, and its diff twin
+  **`updateDiffTabContent(id, original, modified, tick)`** — a `DiffTab` follows the same staleness
+  contract in `DiffPane`. The transient **`changesRequest`** +
   **`requestChangesView(workspaceId, path)`** are a UI deep-link intent (a chat turn-divider asking the
   right panel to surface a file's diff); the panels watch it, scoped by workspace. **`openDoc(tab)`** opens
```
rsolmano: **Nit — stale prose after the deep-link contract change.**

This sentence still says the deep-link asks the right panel to "surface a file's diff", but this PR changed the contract to highlight-only (the diff opens on explicit click — `panels/SPEC.md` records it). Same sweep should catch `apps/web/src/chat/turns.tsx`'s divider comment ("opens those files in the Changes/diff panel"). Specs are ground truth here — worth keeping the wording exact.

rsolmano (commented):
## Review: strict pass over the split-diff viewer (#19)

Verified hands-on, not just read: all gates run locally (`check:deps` / `lint` / `typecheck` ✅, unit 179/179 ✅, `bun run e2e` 47/47 ✅), plus a fresh isolated host driven through the whole feature surface (split/inline, List|Tree, live refresh, deleted/renamed/untracked/binary files), plus targeted `@agent` runs on main, on this branch, and on its local merge with main.

**The feature itself is good work** — the issue is solved, the SPEC.md updates lead the code (panels/store/git/host/web), the wire change is done right (`PROTOCOL_VERSION` 8→9, zero `git.diff`/`DiffViewer` leftovers), and the extraction of `monacoSetup`/`TreeRow`/`ToggleSegment`/`DiffStatBadge` removes duplication rather than adding it. Live refresh, per-tab view state, folder aggregation, tab dedupe: all verified working. No new suppressions, no new deps.

**But it is not mergeable as-is:**

### 🔴 Blocker — the `@agent` e2e suite is now red
`e2e/turn-divider.live.spec.ts` was not migrated: it still asserts the deleted `diff-viewer` testid *and* pins the old deep-link contract (chip → diff shown) that this PR deliberately changes to highlight-only. Proven empirically: the spec **passes on main (7.4s)** and **fails on this PR merged with main** at exactly that assertion. Only the no-agent suite (`bun run e2e`, which grep-inverts `@agent`) was kept green — `bun run e2e:full` was evidently never run. Please rewrite the spec to the new contract recorded in `panels/SPEC.md` (assert the `change-item` highlight; drop `diff-viewer`).

### 🟠 Stale base — rebase required
The branch sits 2 commits behind main (pre-ModelRuntime, pi **0.80.6** vs main's **0.81.1**). It merges cleanly and the *merged* result works, but on this branch's own head the live agent→Changes flow is broken (reproduced: a real agent writes a file, the Changes panel says "No changes in this workspace" — inherited from the stale base, not this PR's code). Rebase onto main and re-run `bun run e2e:full` so what's reviewed is what ships.

### Findings (details on the inline comments)
| Severity | Finding |
|---|---|
| Major | Untracked **binary** files get garbage `+N` badge counts + every untracked file is fully read on every `git.status` (perf/memory) |
| Minor | Two Changes rows can render "selected" at once (stale click/deep-link highlight) |
| Minor | "Split" silently renders inline below ~900px (Monaco `useInlineViewWhenSpaceIsLimited` default) — toggle looks like a no-op |
| Minor | `gitDiffFile` conflates any `git show` failure with "absent in base" → silent whole-file-added diff |
| Refactor | `DiffPane`'s live-refresh effect is a wholesale copy of `FilePane`'s — centralize (repo bar: no duplicated derivations) |
| Nit | Stale prose: store SPEC + `chat/turns.tsx` still describe the old "chip opens the diff" contract |

### PR hygiene
Branch is named `olga.lavrichenko/todos-mockups` for a diff-viewer PR; the body is a bare screenshot — no description, no `Closes #19`, no test plan. Please add these.

### Required before merge
- [ ] Migrate `e2e/turn-divider.live.spec.ts` to the new deep-link contract
- [ ] Rebase onto main; run `bun run e2e:full` and say so in the PR
- [ ] Fix untracked-binary badge counts (and cap the per-file reads)
- [ ] Nice-to-have: dedupe the live-refresh effect, clear the stale highlight, sweep the two stale prose spots, decide the narrow-width Split behavior

rsolmano (approved):

PR #105, opened 2026-07-22
feat: built-in Default workspace + an explicit choice between the two working modes

rsolmano:
> **Stacked on #134** (the copy-only quick win) — this PR targets `welcome-copy-setup-note`, so its diff shows only the Default-workspace work. It retargets to `main` automatically once #134 merges.

## What

People were getting lost in the project → workspace (git worktree) model: opening a project landed on a Welcome screen with zero workspaces, and nothing represented "just work in my project folder."

Every project now carries exactly one built-in **Default workspace** (`Workspace.kind: "default"`) whose cwd is the **project folder itself** (git's *main working tree*) — **non-removable and non-renamable**, enforced server-side, not just hidden in the UI.

And the two working modes are now an **explicit choice** rather than a hidden default (second commit, after testing feedback): opening a project lands on its **Welcome**, where *Start building* (isolated worktree) sits beside *Work in project folder* (enters the Default), and the New-Workspace dialog carries a **target control** with a mode-aware header.

## Decisions (user-confirmed in the design rounds)

- **Label "Default"**, live current branch beneath it (the folder can be on any branch, so the label must not claim "main"), pinned first with a House icon.
- ~~Auto-enter on open~~ → **superseded**: opening a project lands on **Welcome**, the mode fork. Auto-enter hid that a choice existed.
- **Dialog target control** — *Isolated workspace* ↔ *Project folder*, always both visible; folder mode creates nothing (submit **enters** the Default), hides the base-branch picker + naming hint, and the button reads **Start**.
- **"Set up project"** keeps the full dialog, preselects **Project folder** (specs are ground truth — they land in place, no merge dance) and carries the explanatory note (the note itself ships in #134).
- **Changes = diff vs the repo's default branch** (`origin/HEAD` → `origin/main` → repo HEAD), same semantics as worktree workspaces.
- **Invariants:** exactly one per project · non-renamable · non-removable.

## How

- **contracts** — `Workspace.kind?: "default"` (explicit wire field, never an id convention); `PROTOCOL_VERSION` **17**.
- **server/workspaces** — lazily **ensured in `listWorkspaces`** (find-or-create: backfills existing projects, self-heals out-of-band state churn, collapses duplicate records); `branch`/`baseBranch` are folder-truth, refreshed quietly on drift; `forgetWorkspace`/`renameWorkspace` **throw** on it and `reclaimWorktree` refuses it — the archive teardown's `rm -rf` fallback must never see the project folder; `ensureWorkspaceScratchDir` extracted.
- **server/git** — `resolveDefaultBranch` factored (shared by the base picker + the ensure); `currentBranch` (unborn-HEAD-safe); `gitStatus` reports the Default's branch **live** so the Changes header never lags a terminal `git checkout`.
- **server/host** — `workspace.remove` rejects the Default loudly **before any side-effect**; `session.create` seeds the gitignored `.thinkrail/context/` scratch dir — merely opening a project never writes into the user's repo, starting a chat there does (zero git footprint via the self-ignoring gitignore).
- **web** — pinned Default row (House icon, no Remove button, `data-kind` hook); Welcome fork card + dialog target segment; new shared `panels/defaultWorkspace.ts` (`resolveDefaultWorkspace` — one resolve+degrade path for both); truthful empty-center receipt ("Chats, changes, and terminals run directly in your project folder").
- **specs first** — `workspaces`/`git`/`host`/`contracts`/`panels` SPECs, the `architecture.md` decision #6 exception, and `goal-and-requirements.md` updated as part of the change.

## Verification

- `bun run test` — **335 pass** (10 new workspaces tests: ensure idempotence, drift refresh, duplicate collapse, guards incl. `reclaimWorktree` refusing the project folder, scratch-dir laziness; 1 new git test: live status branch).
- `bun run e2e` — **101/101**, incl. `e2e/default-workspace.spec.ts` (fork card → Default, folder-cwd proof via terminal `pwd`, pin-first, no-Remove, uniqueness across re-opens, project-home gesture). Specs that landed on `main` meanwhile (`history-jump`, `history-search`, `templates.live`, `new-workspace`) were migrated to the `worktreeRows` fixture so the always-present Default row can't be mistaken for a worktree.
- `check:deps` / `lint` / `typecheck` clean; no suppressions added. Rebased onto fresh `main` (pi 0.82.1) — protocol 9 → 17, `DiffStatBadge`, slash-command completion and the skills/trust surfaces all reconciled.

### DIFF
--- a/apps/web/src/panels/CenterTabs.tsx
+++ b/apps/web/src/panels/CenterTabs.tsx
@@ -21,6 +21,7 @@ import {
 	type DocTab,
 	type EditorTab,
 	selectActiveWorkspace,
+	selectContextProject,
 	selectWorkspaceTick,
 	toast,
 	useAppStore,
@@ -95,6 +96,7 @@ function ChatHistoryMenu({
 export function CenterTabs() {
 	const activeWorkspaceId = useAppStore((s) => s.activeWorkspaceId);
 	const activeWorkspace = useAppStore(selectActiveWorkspace);
+	const contextProject = useAppStore(selectContextProject);
 	const tabsByWorkspace = useAppStore((s) => s.tabsByWorkspace);
 	const activeTabByWorkspace = useAppStore((s) => s.activeTabByWorkspace);
 	const closedChatsByWorkspace = useAppStore((s) => s.closedChatsByWorkspace);
@@ -238,6 +240,9 @@ export function CenterTabs() {
 		else closeTab(tab.id);
 	};
 
+	// The Default workspace is the project folder itself — its receipt must tell the truth ("runs
+	// directly in your project folder") instead of promising worktree isolation.
+	const isDefaultWorkspace = activeWorkspace?.kind === "default";
 	const placeholder = (
 		<div className="flex h-full flex-col items-center justify-center gap-md px-lg text-center text-hint">
 			{activeWorkspace ? (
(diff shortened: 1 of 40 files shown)


danyaberezun:
I have some doubts about the concept in general as it is here right now. I would say that by default we should somehow encourage a user to use workspaces unless he explicitly deny using them for a project.

PR #103, opened 2026-07-22
feat(chat): Ctrl+R history search + prompt templates

danyaberezun:
**Ctrl+R history search** — recall your prompts + full-text search over conversations, scoped chat → workspace → project → everywhere. Zoomed view with a preview pane and jump-to-message; ↑ inline recall; assistant-only message hits with jumpable prompt rows.

**Prompt templates** — pi-native files (portable to the pi CLI), invoked from the `/` menu with Tab-through placeholder slots (repeated slots mirror; gaps highlighted in the field). Templates manager in Settings, save-as-template from history, starter-pack offer. Ships five starter templates in `.pi/prompts/`.

Wire: `history.search` + `template.*` (protocol → v11). Server: lazy in-memory index over pi session files. Rebased on main (incl. pi 0.81 upgrade). e2e: 46 → 80, all green.

PR #109, opened 2026-07-23
feat(chat): Ctrl+R history search

danyaberezun:
Ctrl+R prompt-history search + recall over pi's session files (like Claude Code's Ctrl+R).

- Lazy in-memory index → `history.search` (protocol **v9**), scoped chat → workspace → project → everywhere; prompt recall + full-text conversation matches, recency-ordered.
- Overlay (Ctrl+R, or the composer's history button): zoomed two-pane preview, scope picker, assistant-only message hits, jumpable prompt rows (⇧⏎ → scroll + flash the anchored turn, cross-workspace).
- Plain ↑/↓ recalls this chat's own prompts.

First of two stacked PRs (splitting the old #103 into search + templates). Rebased on main (pi 0.81). e2e: 63 no-agent specs green.

### DIFF
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -329,10 +329,14 @@ export default function ChatView({
 		const prefix = anchorText.slice(0, 40);
 		const mappedId = runtime.turnIdByMessageIndex?.[messageIndex];
 		const mapped = mappedId ? turns.find((t) => t.id === mappedId) : undefined;
+		// Fall back to the NEWEST turn whose text matches the anchor (`findLast`, not `find`): a hit is
+		// deduped to its newest occurrence, so when there's no exact index map (a never-hydrated live chat)
+		// or the mapped turn's text no longer matches, the newest match is the right target — matching the
+		// oldest would jump repeated prompts to a stale earlier turn.
 		const target =
 			mapped && turnAnchorText(mapped).includes(prefix)
 				? mapped
-				: turns.find((t) => turnAnchorText(t).includes(prefix));
+				: turns.findLast((t) => turnAnchorText(t).includes(prefix));
 		const index = target ? rowIndexForTurn(rows, target.id) : -1;
 		if (index === -1) {
 			toast.error("couldn't locate the message — the session may have changed");
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -307,11 +307,12 @@ export default function ChatView({
 		closeHistory();
 	};
 
-	// Cmd/Ctrl+Enter on a prompt hit: reuse the exact `onSubmit` path a normal composer send takes, then
-	// clear the draft the same way `Composer`'s own `submit()` does after sending.
+	// Cmd/Ctrl+Enter on a prompt hit: send through the composer's own submit seam (its imperative
+	// handle), never a ChatView-side `onSubmit` call — the composer privately holds pending image
+	// attachments, and only its own path sends them with the text and clears them with the draft
+	// (which also resets the store draft via the composer's `onChange("")`).
 	const onInsertAndSendHit = (hit: PromptHit) => {
-		onSubmit(hit.text, [], isStreaming ? "followUp" : "send");
-		useAppStore.getState().setChatDraft(sessionId, "");
+		composerRef.current?.insertAndSubmit(hit.text, isStreaming ? "followUp" : "send");
 		closeHistory();
 	};
 
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -1,6 +1,7 @@
 import type {
 	AskUserQuestionResult,
 	ImageContent,
+	PromptHit,
 	ThinkingLevel,
 	WireModel,
 } from "@thinkrail/contracts";
@@ -8,21 +9,55 @@ import { ArrowDown } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Virtuoso, type VirtuosoHandle } from "react-virtuoso";
 import { Popover, PopoverAnchor, PopoverTrigger } from "@/components/ui/popover";
-import { EMPTY_RUNTIME, selectSkillsStale, useAppStore } from "@/store";
+import { EMPTY_RUNTIME, selectSkillsStale, toast, useAppStore } from "@/store";
 import { errorText, getTransport } from "@/transport";
 import { AskStatesContext, deriveAskStates } from "./askState";
 import { type ChatActions, ChatActionsContext } from "./ChatActions";
 import { ChatHeader } from "./ChatHeader";
 import { ChatPlanContent, ChatPlanStripContent } from "./ChatPlan";
-import { Composer, type MentionCandidate, type SubmitBehavior } from "./Composer";
+import {
+	Composer,
+	type ComposerHandle,
+	type MentionCandidate,
+	type SubmitBehavior,
+} from "./Composer";
 import { ExtUiDialog } from "./ExtUiDialog";
-import { type ChatRow, deriveRows } from "./rows";
+import { HistoryOverlay } from "./HistoryOverlay";
+import { type ChatRow, deriveRows, rowIndexForTurn } from "./rows";
 import { SkillsDialog } from "./SkillsDialog";
 import { StreamIndicator, type StreamStatus, streamStatus } from "./StreamIndicator";
 import "./tools/register"; // side-effect: register the built-in pi tool renderers (bash/read/edit/write)
 import { ChatTurnView } from "./turns";
+import type { ChatTurn } from "./types";
 import { useChatScroll } from "./useChatScroll";
 import { useChatTodos } from "./useChatTodos";
+import { useHistorySearch } from "./useHistorySearch";
+
+/**
+ * Best-effort plain text for a turn — user: `message.content` (a plain string, or text/image blocks);
+ * assistant: joined text blocks. `system`/`error`/`retry` turns fall through to `""`. Two consumers:
+ * anchor-matching a `chatLocationRequest` jump target (`turnIdByMessageIndex` only maps user/assistant
+ * messages, so a non-anchor turn's `""` never matches and is never wrongly selected by the fallback scan),
+ * and `recentPrompts` below (user turns only, so the assistant/`""` branches never surface there).
+ */
+function turnAnchorText(turn: ChatTurn): string {
+	if (turn.kind === "user") {
+		const { content } = turn.message;
+		return typeof content === "string"
+			? content
+			: content
+					.filter((b) => b.type === "text")
+					.map((b) => b.text)
+					.join("\n");
+	}
+	if (turn.kind === "assistant") {
+		return turn.message.content
+			.filter((b) => b.type === "text")
+			.map((b) => b.text)
+			.join("\n");
+	}
+	return "";
+}
 
 /** Context threaded to the Virtuoso footer so the streaming loader lives at the end of the conversation. */
 type ChatListContext = { status: StreamStatus | null };
@@ -57,12 +92,13 @@ export default function ChatView({
 	// chat streaming into its own runtime never re-renders the foreground one.
 	const runtime = useAppStore((s) => s.sessions[sessionId]) ?? EMPTY_RUNTIME;
 	const models = useAppStore((s) => s.models);
-	// This chat's owning project (workspaces are keyed by project) — for the Skills manager's trust ops.
+	// This chat's owning project (workspaces are keyed by project) — for the Skills manager's trust ops
+	// and the "project" / "all" history-search scopes.
 	const projectId = useAppStore(
 		(s) =>
 			Object.values(s.workspaces)
 				.flat()
-				.find((w) => w.id === workspaceId)?.projectId ?? null,
+				.find((w) => w.id === workspaceId)?.projectId,
 	);
 	const [skillsOpen, setSkillsOpen] = useState(false);
 	// Auto-detect: a skill dir changed on disk (pull/branch/edit) after this session loaded — flag a manual
@@ -76,6 +112,17 @@ export default function ChatView({
 		}
 		return undefined;
 	});
+	// The raw record is a stable reference from zustand's perspective (unlike a fresh object/array
+	// literal, which would re-render this view on every unrelated store update); the workspaceId →
+	// display-name map the history overlay's cross-workspace chip needs is derived below in a `useMemo`.
+	const workspaces = useAppStore((s) => s.workspaces);
+	const workspaceNames = useMemo(() => {
+		const map: Record<string, string> = {};
+		for (const list of Object.values(workspaces)) {
+			for (const w of list) map[w.id] = w.name;
+		}
+		return map;
+	}, [workspaces]);
 	const {
 		turns,
 		toolResults,
@@ -108,6 +155,20 @@ export default function ChatView({
 		return { status };
 	}, [turns, isStreaming, currentAssistantId]);
 
+	// The plain `↑`-recall list (`Composer`'s `recentPrompts` prop): this chat's own user-turn texts,
+	// newest first, deduped keeping the NEWEST occurrence — the same recency-first ranking rule as the
+	// server history index (and the atuin/fzf convention it follows). Reuses `turnAnchorText`'s
+	// user-content extraction (string, or joined text blocks) rather than re-deriving it. Reverse to
+	// newest→oldest *before* deduping: `Set` keeps each string's first-seen entry, so reversing first
+	// makes that first-seen entry the newest one, not the oldest (see `chat/SPEC.md`).
+	const recentPrompts = useMemo(() => {
+		const texts = turns
+			.filter((t) => t.kind === "user")
+			.map((t) => turnAnchorText(t))
+			.filter(Boolean);
+		return [...new Set(texts.reverse())];
+	}, [turns]);
+
 	const [mentionQuery, setMentionQuery] = useState<string | null>(null);
 	const [mentionCandidates, setMentionCandidates] = useState<MentionCandidate[]>([]);
 	// The chat's TODO plan, surfaced inline via a header strip that opens a popup over the chat (SPEC §Chat TODO plan).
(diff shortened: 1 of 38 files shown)


rsolmano (commented):
## Comprehensive Code Review

Reviewed commit 16d8a8327c0b43acb368539a9e9960874a3415c2.

### Findings Summary

| Priority | Issue | Location | Review type |
|----------|-------|----------|-------------|
| P1 | Index the canonical active transcript, not physical JSONL order | packages/server/src/history/extract.ts:35 | architecture(anthropic/claude-opus-4-8), manual reproduction |
| P1 | Invalidate old results when the query or scope changes | apps/web/src/chat/useHistorySearch.ts:223 | bugs(openai/gpt-5.6-terra) |
| P1 | Keep project and workspace state atomic for cross-project jumps | apps/web/src/store/appStore.ts:1142 | architecture(openai/gpt-5.6-terra) |
| P1 | Remove the forbidden lint suppression | apps/web/src/chat/HistoryOverlay.tsx:334 | code-quality(manual policy audit) |
| P2 | Do not prefix-match duplicate text in already-live chats | apps/web/src/chat/ChatView.tsx:330 | bugs(openai/gpt-5.6-terra) |
| P2 | Exclude internal TODO nudges from prompt history | packages/server/src/history/extract.ts:63 | requirements-compliance(anthropic/claude-opus-4-8, openai/gpt-5.6-terra) |
| P2 | Reset plain-history recall when inserting a search result | apps/web/src/chat/Composer.tsx:177 | bugs(anthropic/claude-opus-4-8) |
| P2 | Dismiss slash and mention menus from the history button path | apps/web/src/chat/Composer.tsx:423 | code-quality(openai/gpt-5.6-terra) |
| P2 | Avoid reparsing every session during warm revalidation | packages/server/src/history/historyIndex.ts:53 | performance(anthropic/claude-opus-4-8, openai/gpt-5.6-terra), manual benchmark |
| P2 | Validate and cap client-controlled result limits | packages/server/src/history/historyIndex.ts:126 | security(anthropic/claude-opus-4-8, openai/gpt-5.6-terra), performance(openai/gpt-5.6-terra) |
| P2 | Keep E2E fixtures behind an explicit module boundary | e2e/fixtures/sessions.ts:11 | architecture(anthropic/claude-opus-4-8), code-quality(manual policy audit) |

### Recommendation

Request changes: address the four P1 findings before merge. The existing checks are green, but they do not cover branch/compaction-aware indexing, stale-result actions, or cross-project state consistency.

packages/server/src/history/extract.ts
```diff
@@ -0,0 +1,74 @@
...
+		.join("\n");
+}
+
+/** Parse one pi session JSONL file into searchable entries. Tolerant: non-JSON / unknown lines skip
+ * without disturbing the renderable-message count. Pure. */
+export function extractEntries(jsonl: string): HistoryEntry[] {
+	const entries: HistoryEntry[] = [];
+	let messageIndex = 0;
+	for (const raw of jsonl.split("\n")) {
+		const lineText = raw.trim();
```
rsolmano: **[P1] Index the canonical active transcript, not physical JSONL order** (review type: architecture(anthropic/claude-opus-4-8), manual reproduction)

Pi session files are trees. Iterating every physical message line indexes abandoned branches and assigns indices that differ from the array returned by pi's buildSessionContext/getMessages path; compaction creates the same mismatch. I reproduced a branched session where the extractor exposed an abandoned message at index 2 and the active message at index 4, while pi exposed the active message at index 2. These hits can therefore jump to the wrong turn or to content that is not present in the hydrated transcript.

**Suggested fix:** derive entries through pi's exported parse/migrate/context-building functions so the indexed, filtered array exactly matches hydration. Add branch and compaction fixtures. If abandoned branches intentionally remain searchable, make those hits non-jumpable unless navigation becomes branch-aware.

apps/web/src/chat/useHistorySearch.ts
```diff
@@ -0,0 +1,287 @@
...
+		setScopeKind("workspace");
+		setQueryState(seedQuery);
+		setSelected(0);
+		setResult(null);
+		setError(false);
+	}, []);
+
+	const close = useCallback(() => setOpen(false), []);
+
+	const setQuery = useCallback((q: string) => {
```
rsolmano: **[P1] Invalidate old results when the query or scope changes** (review type: bugs(openai/gpt-5.6-terra))

setQuery and the scope setters reset only the selection. The previous result remains rendered and actionable during the 100 ms debounce and subsequent request; the sequence token prevents an old response from landing, but it does not stop click/Enter/Ctrl+Enter from acting on this already-stale result. Ctrl+Enter can consequently submit a prompt that does not match the visible query.

**Suggested fix:** synchronously clear/invalidate result on every query or scope change, or key results to the request inputs and disable selection until that key matches the current query and scope. Add a delayed-response interaction test.

apps/web/src/store/appStore.ts
```diff
@@ -1101,6 +1138,9 @@ export const useAppStore = create<AppState>((set, get) => ({
 	setSettingsSection: (section) => set({ settingsSection: section }),
 	applyConfig: (config) => set({ theme: config.theme }),
 	requestChangesView: (workspaceId, path) => set({ changesRequest: { workspaceId, path } }),
+	requestChatLocation: (req) =>
+		set({ chatLocationRequest: req, activeWorkspaceId: req.workspaceId }),
```
rsolmano: **[P1] Keep project and workspace state atomic for cross-project jumps** (review type: architecture(openai/gpt-5.6-terra))

History hits carry projectId, but ChatLocationRequest drops it and this action changes only activeWorkspaceId. A jump into another project's not-yet-loaded workspace leaves selectedProjectId on the source project and selectActiveWorkspace unresolved; I reproduced selectedProjectId=p1, activeWorkspaceId=w2, activeWorkspace=null after targeting p2. This bypasses the store's existing atomic project/workspace navigation invariant.

**Suggested fix:** carry projectId in the location request, ensure the target project's workspace list is loaded, and activate both IDs atomically. Cover a two-project jump where the destination project is initially unloaded.

apps/web/src/chat/HistoryOverlay.tsx
```diff
@@ -0,0 +1,560 @@
...
+
+	// Arrow-key navigation moves `selected` past the edge of what's currently scrolled into view — the
+	// container itself scrolls (mouse wheel, drag), but a keyboard-only selection change never does on its
+	// own. `block: "nearest"` is the minimal scroll: it only moves the container when the selected row is
+	// actually outside the visible range, never re-centers a row that's already visible. `selected`/
+	// `stage`/`result` are all trigger-only — the body reaches the selected row through the DOM (via
+	// `data-selected`), not through these values directly — but a stage toggle or a fresh result can change
+	// which row `selected`'s index now points at without the index itself changing, so all three must
+	// re-run the effect.
+	// biome-ignore lint/correctness/useExhaustiveDependencies: selected/stage/result are re-render triggers, not body inputs — the row is found via the DOM, not these values
```
rsolmano: **[P1] Remove the forbidden lint suppression** (review type: code-quality(manual policy audit))

This newly added biome-ignore conflicts with the repository's explicit no-suppressions handoff rule. The lint gate is green only because the diagnostic is hidden, whereas the project requires restructuring the state/effect when a dependency warning exposes that mismatch.

**Suggested fix:** derive a stable selection key/item that the effect actually consumes, or use a callback ref for scrolling, so the dependency list is naturally correct without a suppression.

apps/web/src/chat/ChatView.tsx
```diff
@@ -201,6 +292,64 @@ export default function ChatView({
...
+			return;
+		}
+		const { messageIndex, anchorText } = chatLocationRequest;
+		const prefix = anchorText.slice(0, 40);
+		const mappedId = runtime.turnIdByMessageIndex?.[messageIndex];
+		const mapped = mappedId ? turns.find((t) => t.id === mappedId) : undefined;
+		const target =
+			mapped && turnAnchorText(mapped).includes(prefix)
+				? mapped
+				: turns.find((t) => turnAnchorText(t).includes(prefix));
```
rsolmano: **[P2] Do not prefix-match duplicate text in already-live chats** (review type: bugs(openai/gpt-5.6-terra))

Fresh/already-live runtimes have no turnIdByMessageIndex, so this fallback chooses the first turn whose text contains the prefix. Repeated prompts therefore jump to the oldest occurrence even though prompt deduplication keeps the newest occurrence, and repeated assistant text has the same ambiguity.

**Suggested fix:** maintain the exact message-index-to-turn mapping while folding live events instead of resolving by text. Add repeated-prompt and repeated-assistant-message navigation tests.

packages/server/src/history/extract.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		// message entries: require message.role to be a known type.
+		if (entry.type !== "message") continue;
+		const role = entry.message?.role;
+		if (typeof role !== "string" || !MESSAGE_ROLES.has(role)) continue;
+		const index = messageIndex++;
+
+		// Extract text only from user/assistant roles.
+		if (role !== "user" && role !== "assistant") continue;
+		const text = textOf(entry.message?.content);
```
rsolmano: **[P2] Exclude internal TODO nudges from prompt history** (review type: requirements-compliance(anthropic/claude-opus-4-8, openai/gpt-5.6-terra))

Hydration deliberately hides user messages prefixed with [thinkrail:todo-nudge], but this extractor indexes them as ordinary recallable prompts. That exposes internal control traffic in Ctrl+R and allows it to be inserted or resubmitted by the user.

**Suggested fix:** centralize the internal-message marker and skip these entries after incrementing the canonical message index, preserving anchor alignment. Add an extraction/search regression test.

apps/web/src/chat/Composer.tsx
```diff
@@ -117,15 +147,42 @@ export function Composer({
...
+		}
+		setCaret(pendingCaret);
+		setPendingCaret(null);
+	}, [pendingCaret]);
+
+	/** Place a collapsed caret at `pos` (after the current commit — see `pendingCaret` above). */
+	const focusCaret = useCallback((pos: number) => setPendingCaret(pos), []);
+
+	useImperativeHandle(
+		handleRef,
```
rsolmano: **[P2] Reset plain-history recall when inserting a search result** (review type: bugs(anthropic/claude-opus-4-8))

insertText replaces the draft but leaves recallIdx active. After ArrowUp starts a recall session, choosing a Ctrl+R result and then pressing ArrowDown can clear or overwrite the newly inserted prompt.

**Suggested fix:** reset recallIdx whenever a non-recall action replaces the draft. Prefer one helper for programmatic draft replacement so history, mention, and slash insertion follow the same rule.

apps/web/src/chat/Composer.tsx
```diff
@@ -311,6 +414,17 @@ export function Composer({
 						<ThinkingSelector level={thinkingLevel} onSelect={onSelectThinking} />
 					</div>
 					<div className="flex shrink-0 items-center gap-sm">
+						{/* Always rendered — the tap path to history recall on mobile, and a discoverability
+						 * affordance for `Ctrl+R` on desktop; both open the exact same overlay via `onHistoryOpen`. */}
+						<button
+							type="button"
+							data-testid="history-open"
+							aria-label="Search history"
+							onClick={() => onHistoryOpen?.()}
```
rsolmano: **[P2] Dismiss slash and mention menus from the history button path** (review type: code-quality(openai/gpt-5.6-terra))

The Ctrl+R path calls setDismissed(true), but this button invokes onHistoryOpen directly. If slash or mention suggestions are open, clicking the button can leave both floating panels active over the same anchor.

**Suggested fix:** route keyboard and button activation through one openHistory function that dismisses competing menus before opening the history overlay.

packages/server/src/history/historyIndex.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export class HistoryIndex {
+	private records = new Map<string, SessionRecord>(); // keyed by file path
+	private building: Promise<void> | null = null;
+	private built = false;
+	private lastCheck = 0;
+
+	constructor(private sessionDir?: string) {}
+
+	private async listInfos() {
```
rsolmano: **[P2] Avoid reparsing every session during warm revalidation** (review type: performance(anthropic/claude-opus-4-8, openai/gpt-5.6-terra), manual benchmark)

SessionManager.listAll is not a cheap file listing: pi reads/parses every JSONL file to build SessionInfo. Every revalidation therefore scans the full corpus before this code stats files and rereads changed ones, so the stated changed-files-only warm path is not actually bounded by the mtime diff. On a 134-file/39 MB corpus, a warm revalidation took about 79 ms and this cost scales with total history.

**Suggested fix:** discover files through cheap directory enumeration/stat or a watcher and parse only new/changed files; keep the discovery and reads asynchronous so searches do not wait on a full-corpus parse.

packages/server/src/history/historyIndex.ts
```diff
@@ -0,0 +1,199 @@
...
+	}): Promise<HistorySearchResult> {
+		const wasCold = !this.built;
+		this.ensureFresh();
+		if (this.building) {
+			if (wasCold) await Promise.race([this.building, Bun.sleep(COLD_BUILD_BUDGET_MS)]);
+			else await this.building;
+		}
+		const indexing = !this.built;
+
+		const limit = input.limit ?? 50;
```
rsolmano: **[P2] Validate and cap client-controlled result limits** (review type: security(anthropic/claude-opus-4-8, openai/gpt-5.6-terra), performance(openai/gpt-5.6-terra))

The raw client value reaches slice unchanged. Huge values defeat the protocol's advertised cap, while negative values have surprising semantics; for example, limit=-1 returns every item except the last one.

**Suggested fix:** define a protocol maximum and clamp a finite integer limit at both the request boundary and HistoryIndex.search as defense in depth. Consider bounding query length for the same reason.

e2e/fixtures/sessions.ts
```diff
@@ -0,0 +1,86 @@
...
+// has something to find. Reuses `writeFixtureSession`/`defaultSessionDirFor` from
+// `packages/server/src/history/testFixtures.ts` (a cross-package source import — this file runs under
+// node via Playwright's own TS loading, same as the rest of `global-setup.ts`, and `testFixtures.ts` only
+// imports `node:fs`/`node:path`, so this resolves cleanly with no bundler involved). Those two helpers are
+// pinned against pi's real `SessionManager` by `packages/server/src/history/testFixtures.test.ts` — if
+// pi's on-disk format or default-layout encoding ever shifts, that test fails, not this seeder silently.
+import {
+	defaultSessionDirFor,
+	writeFixtureSession,
+} from "../../packages/server/src/history/testFixtures";
```
rsolmano: **[P2] Keep E2E fixtures behind an explicit module boundary** (review type: architecture(anthropic/claude-opus-4-8), code-quality(manual policy audit))

This cross-package source import bypasses the server package's public surface, while the production history barrel separately re-exports disk-writing test helpers. That conflicts with ThinkRail's top-priority fractal module-boundary rule and pulls test support into the runtime module graph.

**Suggested fix:** expose an explicit test-only package subpath/export, or keep the fixture implementation inside e2e. Remove test helpers from the production history barrel and document the sanctioned test boundary.

packages/server/src/history/historyIndex.ts
```diff
@@ -0,0 +1,234 @@
...
+export class HistoryIndex {
+	private records = new Map<string, SessionRecord>(); // keyed by file path
+	private building: Promise<void> | null = null;
+	private built = false;
+	private lastCheck = 0;
+
+	constructor(private sessionDir?: string) {}
+
+	private async listInfos() {
+		return this.sessionDir ? SessionManager.listAll(this.sessionDir) : SessionManager.listAll();
```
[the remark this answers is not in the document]
danyaberezun: Fixed in 68286ab. `HistoryIndex` no longer calls `SessionManager.listAll()` anywhere: it enumerates `.jsonl` files itself with async `readdir`/`stat` (replicating pi's pinned discovery layout — flat for a custom `sessionDir`, one level of per-cwd subdirs under `<agentDir>/sessions` for the default, agent dir resolved live via pi's exported `getAgentDir()`), and only new/changed files (by mtime+size) are read — once, with async `readFile` — and re-parsed. Session identity (id/cwd/title) now comes from that same single parse (`extractSession`), so nothing reads a file twice. An unchanged corpus revalidates for a readdir walk plus one stat per file (no reads, no parses), and every file boundary in the walk has an await, so the longest uninterrupted event-loop stretch is one file's parse. Pinned by a new unit test driving a no-arg index against the real default layout via `PI_CODING_AGENT_DIR`.

packages/server/src/history/extract.ts
```diff
@@ -0,0 +1,84 @@
...
+		const index = messageIndex++;
+		if (message.role !== "user" && message.role !== "assistant") continue;
+		const text = textOf(message.content);
+		if (!text.trim()) continue;
+		// Internal control traffic: the pi-todos wake-nudge is hidden from the transcript on hydrate, so it
+		// must not surface as a recallable/insertable prompt. The index was already consumed above, so
+		// skipping only the text keeps every later hit's anchor aligned.
+		if (message.role === "user" && text.startsWith(TODO_NUDGE_PREFIX)) continue;
+		out.push({
+			text: text.slice(0, MAX_SEARCHABLE),
```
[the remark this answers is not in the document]
danyaberezun: Fixed in 68286ab. The truncation is removed entirely (`MAX_SEARCHABLE` is gone): `extractSession` stores the full message text, so recall inserts the complete prompt, the overlay's preview really shows the whole thing, and terms past the old 4k cutoff match. For the memory bound we followed pi's own precedent — `SessionInfo.allMessagesText` already holds every session's full text in memory for pi's session picker — rather than adding a separate bounded search representation. Pinned by unit tests at both layers: the extractor round-trips a 10k prompt verbatim, and an index search finds a needle at the very tail of a ~9k prompt and returns the full text on the hit.

apps/web/src/chat/ChatView.tsx
```diff
@@ -220,6 +303,68 @@ export default function ChatView({
...
+	// Enter on a prompt hit: replace the draft, focus, caret at end, close — no submit.
+	const onInsertHit = (hit: PromptHit) => {
+		composerRef.current?.insertText(hit.text);
+		closeHistory();
+	};
+
+	// Cmd/Ctrl+Enter on a prompt hit: reuse the exact `onSubmit` path a normal composer send takes, then
+	// clear the draft the same way `Composer`'s own `submit()` does after sending.
+	const onInsertAndSendHit = (hit: PromptHit) => {
+		onSubmit(hit.text, [], isStreaming ? "followUp" : "send");
```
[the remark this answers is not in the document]
danyaberezun: Fixed in 9e2a220. `ComposerHandle` gains `insertAndSubmit(text, behavior)`, and the overlay's Cmd/Ctrl+Enter now goes through it instead of a ChatView-side `onSubmit` call. Inside the composer, both the keyboard send and `insertAndSubmit` land on the same `submitText` seam, so pending images are sent with the text and cleared with the draft atomically, whichever side initiated the send. Added the requested regression e2e: it pastes an image, insert-and-sends a recalled prompt, and asserts on the actual WebSocket frame that the `session.prompt` request carries the image (plus that the thumbnail strip clears — no stale attachment left for the next message).

rsolmano (dismissed):

PR #128, opened 2026-07-26
feat(catalog-freshness): a live model catalog

rsolmano:
> The stack is gone: **#125** (pi bump) and **#127** (per-model thinking levels) are both merged, and this branch is rebased onto current `main` with only its own commit left. Reviewable standalone.

## What

**PROTOCOL_VERSION 18 → 19** — the new wire method is its own protocol change (18 went to the built-in Default workspace on `main` while this sat in review).

The picker's model list is whatever the host held when the app booted. A model published after that never appears until restart. This makes the catalog live, and makes "Refresh" actually mean something.

**`model.refresh`** (new wire method) awaits the host's catalog refresh and returns the post-refresh list, where `model.list` fires a detached one and serves the current snapshot. `chat/useModelCatalog` becomes the one seam both pickers share; `ModelSelector` gains a "Refresh catalog" footer row.

## Two findings worth your attention

**1. Without `force`, the Refresh button fetches nothing.** pi's `remote-catalog-provider.js` sets `REMOTE_CATALOG_REFRESH_INTERVAL_MS = 4 * 60 * 60 * 1000` and returns early inside that window *before* its `If-None-Match` request — so there isn't even a 304. Measured against the real runtime:

```
refresh({allowNetwork:true})              1ms    ← no egress at all
refresh({allowNetwork:true, force:true})  329ms  ← actual revalidation
```

`force` is now the caller's intent: the Refresh row forces, an incidental popover-open does not. Single-flight is keyed with the kind, so a forced caller never joins a throttled pass and inherits its no-op.

**2. Opening the picker must not block on the network.** Driving the real app with an `ollama` provider configured but not running (an ordinary setup), the awaited refresh on popover-open spun the row for **13.3 seconds** before the host's 15s abort. An open is incidental, so it now takes the snapshot read; only the Refresh row awaits. Re-measured: open never sets `data-refreshing`; the row click spins at 29ms.

## The dialog's selection has to track a list that moves

A live catalog can replace the list underneath a held selection, so the dialog reconciles `{model, effort}` against it via the pure `reconcileModel`. Whether a *missing* model may be **replaced** turns on `catalogFresh`:

- **fresh** (this opening's own `model.list`, or an awaited forced refresh) — `model.list` and `resolveWireModel` read the **same host registry**, so a list fetched now rules exactly as `create()` will: the model really is gone, substitute rather than let Create fail.
- **not fresh** (the app-wide store copy, which can predate the host's last detached refresh while `model.default` answers from the newer one) — confirm-only, or a stale cache would override a valid host default and make Create submit a ref the host rejects.

**Only an awaited forced refresh establishes freshness.** `model.list` never does, however recently it ran: its handler starts a detached refresh and replies from *before* it, so the registry can move underneath the response with the client none the wiser — which is exactly how a removed model would get confirmed as present and reach `create()`.

Mutation-verified in both directions — always-substitute fails the stale case, never-substitute fails the refreshed one.

## Also

The chat's effort picker resolves its model through `selectCatalogModel`, so a refresh that changes what a model supports moves its disabled rows too — previously only the dialog reconciled.

`refreshAvailableModels` is pinned against a refresh that lands *after* the call (off `PI_OFFLINE`, which makes awaited and detached indistinguishable); dropping the `await` previously left the whole suite green.

The review trail from the original omnibus PR (#125) is folded into this one commit rather than left as eight `fix:` commits.

## Gates

`check:deps` ✓ · `check:seams` ✓ · `lint` ✓ · `typecheck` (10 pkgs) ✓ · `test` **359 pass** ✓ · `e2e` no-agent **115/115** ✓ — all re-run after the rebase onto current `main`

Binary suite was **90/90** before the split and the rebases onto current `main`; not re-run since — CI covers it.


apps/web/src/chat/useModelCatalog.ts
```diff
@@ -0,0 +1,89 @@
...
+		if (!active) {
+			setFresh(false);
+			return;
+		}
+		let cancelled = false;
+		// Only a read that actually landed makes the store's copy this activation's own snapshot. A
+		// failed one leaves whatever was already there — which is exactly the stale copy `fresh` exists
+		// to distinguish, so claiming freshness on failure would restore the bug it guards against.
+		void readModels().then((landed) => {
+			if (landed && !cancelled) setFresh(true);
```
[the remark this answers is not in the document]
rsolmano: Both findings agreed and fixed in `b6edad8`.

**Protocol version — my error from splitting the PRs.** I split them specifically so each would own one bump, then didn't bump the second, so this advertised the base's v17 while adding a method the base doesn't have. Now **v18**, with `model.refresh` recorded under its own v18 history entry and v17 left describing the base's `thinkingLevels` change.

**Freshness — you're right that `model.list` can never establish it.** I'd been treating "re-read since this consumer activated" as authoritative, but `listAvailableModels` fires `refreshCatalogsDetached` *before* returning its snapshot, so the registry can move underneath the reply with the client none the wiser. That's precisely how a removed model gets confirmed as present and reaches `create()`.

`fresh` is now set **only** when an awaited forced `model.refresh` result is installed — the one path where the pass it triggered has actually finished. `model.list` still runs on activation to keep the picker's list current, but it no longer claims authority, and the reason is written where the flag is declared so it doesn't get re-broken.

That also resolves the second half of your note: because the flag is set from the refresh result rather than from an activation read, a successful forced refresh can now reconcile a removed selection even if the initial list failed.

**One related simplification while I was in there.** `reconcileSelection` decided model *and* effort together; since #127 moved effort onto the host's clamp, it's now `reconcileModel` and decides only which model. Its tests name the two states explicitly — non-authoritative never substitutes, authoritative does — so the distinction you raised is what the test file is about rather than an implicit boolean.

rsolmano:
Heads-up on CI for this PR: **it gets none.** `ci.yml` triggers on `pull_request: branches: [main]`, and this targets `feat/per-model-thinking-levels`, so no workflow runs while it is stacked. It will pick up CI automatically once #127 merges and this retargets to `main`.

Meanwhile it carries the same e2e fix as #127 (the no-agent runner has no provider auth, so no model resolves and the effort pill is disabled), plus this PR's own Refresh-row assertions. Verified locally in both shapes: with auth **99/99**, and with `PI_CODING_AGENT_DIR` pointed at an empty dir to reproduce the CI condition **99/99**. Typecheck and 341 unit tests green.

apps/web/src/chat/useModelCatalog.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * `model.list`: the host's current snapshot, served without touching the network. Note it is a *moving*
+ * answer — the handler fires its own detached catalog refresh before replying — so it keeps the picker's
+ * list current but never establishes freshness (see `fresh`).
+ */
+function readModels(): Promise<void> {
+	return getTransport()
+		.request("model.list", {})
+		.then((m) => useAppStore.getState().setModels(m))
```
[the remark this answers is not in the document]
rsolmano: Agreed and fixed in `74ee54b`.

You were right about the shape of the mistake: I had `fresh` as **local hook state** while `models` is **app-wide**, so the flag outlived the thing it described. After a forced refresh marked the dialog fresh, any later `model.list` install — this picker reopening, or another `useModelCatalog` consumer (a chat mounting) — replaced the shared list while the flag stayed true, and `reconcileModel` would then substitute against a non-authoritative snapshot. That is precisely the removed-model-reaches-`create()` path the flag exists to prevent, so the guard was defeating itself.

Provenance now lives **beside the list it describes**, as you suggested:

- `store.modelsFresh` — written `true` only by `finishModelsRefresh` *with an installed list*, and dropped by `setModels` in the **same write**, so authority falls with whatever list any consumer replaced.
- `useModelCatalog` **reads** it rather than holding one, which also let its `!active` reset go: the activation `model.list` already invalidates, so per-open freshness is now a consequence of the invariant instead of a second mechanism agreeing with it.
- A failed refresh installs nothing, so it changes neither the list nor its provenance.

`appStore.test.ts` pins all four transitions — list → not fresh, awaited refresh → fresh, **list-after-refresh → dropped** (your race, and the one that fails on the old local-state design), failed refresh → unchanged.

Rebased onto current `main` while I was here: the stack is gone (#125 and #127 both merged), so this is one commit plus the fix, and the protocol bump moved **18 → 19** since `main` took 18 for the built-in Default workspace.

apps/web/src/panels/SPEC.md
```diff
@@ -153,11 +153,16 @@ a project picker, the prompt hero, and the reused
...
   time). The pickers' popovers portal into the dialog node (so their lists scroll under the Dialog scroll
-  lock). Pre-session there is no pi to clamp the effort, so an explicit model switch moves it onto the
-  new model's own set by asking the host for pi's clamp (**`model.clampThinking`**) rather than deciding
-  locally — one effect owns "the held effort must be runnable by the held model", so an explicit switch
-  and a catalog refresh that shrank a model's set both resolve the same way pi would. `model.default` needs no adjustment: the host
-  already returns a self-consistent pair.
+  lock). Their catalog is the shared one — `chat/useModelCatalog`, so the dialog and the chat composer
+  cannot drift — which means it is **live**: the picker's Refresh row can replace the list underneath a
+  held selection. The dialog therefore reconciles the held model against it on every change via the pure
+  **`reconcileSelection`**: re-point to the same `{provider,id}` (the refreshed object, whose
```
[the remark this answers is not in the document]
rsolmano: Fixed in `74ee54b`. `panels/SPEC.md` now names **`reconcileModel`** (model only — effort is the host's clamp, which the same paragraph already describes) and the allowed-dep list drops `effortForModel`, which no longer exists; the paragraph also records that `catalogFresh` is the store's `modelsFresh`, dropped by the next `model.list` install from any consumer.

apps/web/src/chat/useModelCatalog.ts
```diff
@@ -0,0 +1,73 @@
...
+	 * app-wide, so a `model.list` install from anywhere — this picker reopening, another chat mounting —
+	 * replaces the list and drops its authority in the same write. See `store`'s `modelsFresh` for why
+	 * `model.list` can never establish it.
+	 */
+	const fresh = useAppStore((s) => s.modelsFresh);
+
+	useEffect(() => {
+		if (!active) return;
+		void readModels();
+	}, [active]);
```
danyaberezun, line 43: `modelsFresh` is dropped only when the `model.list` **reply** lands, so a flag set by a previous consumer straddles this activation. `NewWorkspaceDialog`'s reconcile effect reads it on first render and can substitute `models[0]` off a stale list — the exact case the flag exists to prevent.

Reachable: refresh in a composer picker → log a provider in Settings (no `model.list`, flag survives, registry moves) → open New Workspace with `model.default` replying before `model.list`.

Drop it synchronously. `NewWorkspaceDialog.tsx:217`'s "this opening's own snapshot" only becomes true with this.

```suggestion
	useEffect(() => {
		if (!active) return;
		// Synchronously, not on the reply: `modelsFresh` set by a *previous* consumer must not straddle
		// this activation, or a reconcile on first render reads a stale list as this opening's own truth.
		useAppStore.getState().dropModelsFreshness();
		void readModels();
	}, [active]);
```

apps/web/src/store/appStore.ts
```diff
@@ -643,6 +658,10 @@ interface AppState {
 	handlePiEvent: (event: PiEvent, sessionId: string) => void;
 	setModels: (models: WireModel[]) => void;
 	bumpTemplatesVersion: () => void;
+	/** Atomic begin/finish of the awaited catalog refresh — `finish` lands the new list (null = failed
+	 * refresh: keep the current list, and with it its provenance) and clears the flag in ONE write. */
+	beginModelsRefresh: () => void;
+	finishModelsRefresh: (models: WireModel[] | null) => void;
```
danyaberezun: Declaration for the above.

```suggestion
	beginModelsRefresh: () => void;
	finishModelsRefresh: (models: WireModel[] | null) => void;
	/** Give up authority without replacing the list — a consumer activating can't yet know whether the
	 * list it inherited still matches the host registry. */
	dropModelsFreshness: () => void;
```

apps/web/src/store/appStore.ts
```diff
@@ -1296,8 +1317,17 @@ export const useAppStore = create<AppState>((set, get) => ({
 	// The event→store dispatcher: route each pi event to its session's runtime, so chats stream independently.
 	handlePiEvent: (event, sessionId) =>
 		set((s) => withRuntime(s, sessionId, (rt) => reduceSessionEvent(rt, event))),
-	setModels: (models) => set({ models }),
+	// A `model.list` snapshot: current, but never authoritative — installing it drops `modelsFresh`.
+	setModels: (models) => set({ models, modelsFresh: false }),
 	bumpTemplatesVersion: () => set((s) => ({ templatesVersion: s.templatesVersion + 1 })),
+	beginModelsRefresh: () => set({ modelsRefreshing: true }),
```
danyaberezun, line 1332:
```suggestion
	beginModelsRefresh: () => set({ modelsRefreshing: true }),
	dropModelsFreshness: () => set({ modelsFresh: false }),
```

packages/server/src/agent/piRuntime.ts
```diff
@@ -62,31 +62,65 @@ export type CatalogRefreshRuntime = Pick<ModelRuntime, "refresh">;
...
+		? existing.task.then(() => runCatalogRefresh(runtime, force))
+		: runCatalogRefresh(runtime, force);
+	const task: Promise<void> = started.finally(() => {
+		if (inflightCatalogRefresh.get(runtime)?.task === task) {
+			inflightCatalogRefresh.delete(runtime);
+		}
+	});
+	inflightCatalogRefresh.set(runtime, { task, force });
+	return task;
+}
```
danyaberezun, line 179: An awaited `model.refresh` has no ceiling. `runCatalogRefresh`'s signal covers only `models.refresh` — pi calls `forceRefreshAvailability()` after it with **no signal** (`model-runtime.js:377`), and that's the per-provider `checkAuth` fan-out where the ollama stall lives. On top of that the primary flow *always* queues: popover-open fires `model.list` → detached pass, then the Refresh click queues behind it, so two unbounded passes serialize.

With `modelsRefreshing` app-wide and no client timeout or cancel, one slow pass leaves "Updating catalog…" disabled in every picker. That's the 13.3s stall re-inherited by the path that awaits.

Bound what *callers* await; single-flight keeps tracking the unbounded pass, so a timed-out caller can't start a second concurrent refresh.

```suggestion
export function refreshCatalogs(
	runtime: CatalogRefreshRuntime,
	{ force = false }: { force?: boolean } = {},
): Promise<void> {
	if (process.env.PI_OFFLINE) return Promise.resolve();
	const existing = inflightCatalogRefresh.get(runtime);
	if (existing && (existing.force || !force)) return withDeadline(existing.task);
	// Free slot ⇒ start now (an implicit trigger must reach pi in the same tick it fires). Otherwise
	// this is a forced caller behind a throttled pass: queue, so pi's own per-provider in-flight dedupe
	// has cleared by the time the forced fetch runs.
	const started = existing
		? existing.task.then(() => runCatalogRefresh(runtime, force))
		: runCatalogRefresh(runtime, force);
	const task: Promise<void> = started.finally(() => {
		if (inflightCatalogRefresh.get(runtime)?.task === task) {
			inflightCatalogRefresh.delete(runtime);
		}
	});
	inflightCatalogRefresh.set(runtime, { task, force });
	return withDeadline(task);
}

/**
 * The ceiling every *caller* gets, started at call time. `runCatalogRefresh`'s abort signal covers only
 * `models.refresh` — pi calls `forceRefreshAvailability()` after it with no signal — and a queued forced
 * pass adds a second pass on top of the wait, so an awaited `model.refresh` is otherwise unbounded and
 * the picker's refresh row spins with no cap. Single-flight still tracks the *unbounded* task, so a
 * timed-out caller can't start a second concurrent refresh; it just serves the registry as it stands.
 */
function withDeadline(task: Promise<void>): Promise<void> {
	return new Promise<void>((resolve) => {
		const timer = setTimeout(() => {
			console.warn(
				`model catalog refresh exceeded ${CATALOG_REFRESH_TIMEOUT_MS}ms; serving cached catalogs`,
			);
			resolve();
		}, CATALOG_REFRESH_TIMEOUT_MS);
		timer.unref?.();
		void task.then(() => {
			clearTimeout(timer);
			resolve();
		});
	});
}
```

packages/server/src/agent/piRuntime.ts
```diff
@@ -62,31 +62,65 @@ export type CatalogRefreshRuntime = Pick<ModelRuntime, "refresh">;
...
+ * own `/model`. Two kinds of caller:
  *
- * Never awaited, never throws: pi's provider freshness throttle decides whether anything is fetched
- * (no `force`), failures are logged and swallowed, and `PI_OFFLINE` (pi's env convention, also set by
- * the e2e harness for hermeticity) disables it entirely.
+ * - **throttled** (default): pi's provider freshness window decides whether anything is fetched. pi
+ *   returns early inside that window — *before* its `If-None-Match` revalidation — so an implicit
+ *   trigger (`model.list`, opening the picker) costs nothing and usually changes nothing.
+ * - **`force`**: bypasses that window, so a user-initiated refresh actually reaches pi.dev (one
+ *   conditional request per credentialed provider; a 304 is the common answer).
```
danyaberezun: pi sends an **unconditional** GET (`remote-catalog-provider.js:58` — `accept` + `User-Agent` only) and compares the `last-modified` header. There is no `If-None-Match` and no 304. The conclusion holds; the stated mechanism doesn't.

```suggestion
 * - **throttled** (default): pi's provider freshness window decides whether anything is fetched. Inside
 *   that window pi returns early *before it issues any request at all* (its fetch is unconditional — no
 *   `ETag`/304, just a `last-modified` comparison on the response), so an implicit trigger
 *   (`model.list`, opening the picker) costs nothing and usually changes nothing.
 * - **`force`**: bypasses that window, so a user-initiated refresh actually reaches pi.dev (one full
 *   catalog GET per credentialed provider).
```

packages/server/src/agent/SPEC.md
```diff
@@ -32,13 +32,20 @@ answer-injection path, and the **restart repair** that keeps re-opened transcrip
...
+    live catalogs is the single-flighted **`refreshCatalogs(runtime)`** (issue #98, mirroring pi's own
+    `/model`) behind two triggers: **detached** via `refreshCatalogsDetached` from `model.list` only
+    (`listAvailableModels` fires it, then serves the current snapshot — the picker read never awaits the
+    network; broader triggers — `model.default`, host boot — were considered and declined) and
+    **awaited** via `model.refresh` (`refreshAvailableModels`, the picker's freshness affordance: await
+    the refresh, then serve the post-refresh snapshot). Per-call `refresh({ allowNetwork: true, force })`,
+    where **`force` is the caller's intent, not a constant**: an *implicit* trigger (`model.list`, opening
+    the picker) leaves it off and pi's **4h provider freshness throttle** decides whether anything is
+    fetched, while a *user-initiated* refresh (the picker's Refresh row → `model.refresh({force:true})`)
+    bypasses it — pi returns early inside that window **before** its `If-None-Match` revalidation, so an
```
danyaberezun: Same as `piRuntime.ts:82` — no conditional request exists.

```suggestion
    bypasses it — inside that window pi returns early **before issuing any request at all** (its catalog
```

packages/server/src/agent/piRuntime.test.ts
```diff
@@ -115,16 +116,45 @@ function fakeRuntime() {
...
 	expect(calls.length).toBe(1);
 	const options = calls[0];
 	expect(options?.allowNetwork).toBe(true);
-	expect(options?.force).toBeUndefined();
+	expect(options?.force).toBe(false);
 	expect(options?.signal).toBeInstanceOf(AbortSignal);
 });
 
+// pi returns early inside its 4h freshness window *before* the If-None-Match revalidation, so without
+// `force` a user-initiated "Refresh catalog" fetches nothing at all. These pin the bypass reaching pi.
```
danyaberezun: Same.

```suggestion
// Inside its 4h freshness window pi returns early before issuing any request at all, so without `force`
// a user-initiated "Refresh catalog" fetches nothing. These pin the bypass reaching pi.
```

apps/web/src/chat/SPEC.md
```diff
@@ -485,8 +495,16 @@ from their `toolCall` args and reply through **`ChatActions`** (see below). Work
...
+  models-catalog seam `panels/NewWorkspaceDialog` also imports per-file, so the two pickers cannot
+  drift; it re-reads `model.list` whenever a consumer activates to keep the list current, and reports
+  **`fresh`** — read straight off the store's `modelsFresh`, because catalog authority belongs to the
+  **shared list**, not to a consumer: true only for the installed result of an awaited forced refresh, and
+  dropped by the next `model.list` install from *any* consumer. `model.list` answers from *before* the
+  detached refresh it triggers, so it is never a basis for concluding a model is gone);
+  `react-markdown` / `remark-gfm` / `shiki` (via `lib/highlighter`); `mermaid`
   (**lazy, `tools/visualize` only**); `react-virtuoso`; `lucide-react`; `components/ui`; `lib`.
 - **Forbidden:** value-importing any `pi` package; a **presentational** renderer importing
   `store`/`transport` (only `ChatView`, its integration hooks, and `TemplateEditorDialog` may — keep the
```
danyaberezun: Half-swept: the Allowed line above was just corrected to include `SkillsDialog.tsx`, this still enumerates the old set. `SkillsDialog` is neither `ChatView` nor a hook. Point at the list instead of restating it.

```suggestion
- **Forbidden:** value-importing any `pi` package; a **presentational** renderer importing
  `store`/`transport` (only the app-integration files enumerated above may — keep the
```

danyaberezun (changes requested):
Four blockers. Verified against pi 0.82.1 in `node_modules`; gates re-run locally (typecheck, 359 unit, e2e 115/115 — all green, so none of these are gate-visible).

1. `modelsFresh` can be inherited across consumer activations, enabling the substitution it exists to prevent.
2. The awaited `model.refresh` has no ceiling — the 15s budget doesn't cover `forceRefreshAvailability()`, and the Refresh click always queues behind the popover-open pass.
3. `If-None-Match`/304 doesn't exist in pi's catalog provider — asserted in three places incl. a SPEC.
4. `chat/SPEC.md` Forbidden line left half-migrated.

Non-blocking, separate: `reconcileModel`'s `models[0]` re-derives `getDefaultModel`'s `pinned ?? available[0]` policy client-side (the same file refuses to do that for effort — `model.default` would be authoritative *and* return a consistent level); dropping the `models.length > 0` guard makes every chat-tab switch a `model.list` → full host `runtime.refresh()`; `new-workspace.spec.ts:83` never observes `data-refreshing="true"`, so dropping `beginModelsRefresh` keeps it green.

apps/web/src/chat/useModelCatalog.ts
rsolmano: Agreed — fixed in 6730a14 exactly as suggested: `dropModelsFreshness` on the store, called synchronously in the activation effect (before the read, not on its reply), plus a store test that pins "authority given up without replacing the list".

While here, the activation no longer fires `model.list` unconditionally either (your non-blocking point): it reads only when the shared list is *empty*, so a chat-tab switch stops hanging a full host `runtime.refresh()` — the Refresh row stays the currency path.

packages/server/src/agent/piRuntime.ts
rsolmano: Confirmed and fixed in 6730a14 with your `withDeadline`. Verified the mechanism in the pinned 0.82.1: `model-runtime.js` `refresh()` awaits `forceRefreshAvailability()` with no signal, which fans out through pi-ai `models.getAvailable()` → `checkProviderAuth` → `apiKey.check` per provider — no timeout anywhere on that leg. Added a unit test (bun fake timers) that a caller's await resolves at the ceiling while the pass it started is still pending, *and* that single-flight still owns that unbounded pass so the timed-out caller can't start a second one. `SPEC.md` now says the 15s budget is applied twice (abort signal + caller ceiling) and why the signal alone isn't enough.

packages/server/src/agent/piRuntime.ts
rsolmano: Pushing back on the mechanism here: `If-None-Match`/304 **does** exist in the version we pin. That was true of 0.81.x/0.82.0, but the catalog `0.82.1` in `bun.lock` has:

```js
// dist/core/remote-catalog-provider.js
// Only revalidate when a cached body backs the validator, so a 304 can never leave the overlay empty.
const validator = stored?.models.length ? stored.etag : undefined;
const response = await fetch(url, {
  headers: { accept: "application/json", "User-Agent": getPiUserAgent(VERSION),
             ...(validator ? { "if-none-match": validator } : {}) },
  signal: context.signal,
});
…
if (response.status === 304 && stored) { await context.store.write({ ...stored, checkedAt }); return; }
```

An `accept` + `User-Agent`-only GET is what an *older* install shows — mine at `~/projects/jetbrains/thinkrail` is still on 0.81.1, which matches your quote exactly, so I suspect that's what got read.

Your underlying point still lands though, so I took it: the early return happens **before any request is issued at all**, and the old wording ("before its `If-None-Match` revalidation") made that sound like a request was in flight. All three places now say "before issuing any request at all (its `If-None-Match` revalidation included)". The one genuinely imprecise clause — "a 304 is the common answer" — is now qualified: conditional, hence usually a 304, *wherever a cached body backs the stored `ETag`* (the first fetch for a provider has no validator).

packages/server/src/agent/SPEC.md
rsolmano: See the reply on `piRuntime.ts:85`: 0.82.1 does send `if-none-match` and handle 304, so the mechanism stays — but the wording now says pi returns early **before issuing any request at all** (its `If-None-Match` revalidation included), which was your real point. Updated here too, in 6730a14.

packages/server/src/agent/piRuntime.test.ts
rsolmano: See the reply on `piRuntime.ts:85`: 0.82.1 does send `if-none-match` and handle 304, so the mechanism stays — but the wording now says pi returns early **before issuing any request at all** (its `If-None-Match` revalidation included), which was your real point. Updated here too, in 6730a14.

apps/web/src/chat/SPEC.md
rsolmano: Fixed as suggested in 6730a14 — the Forbidden line now points at the enumerated Allowed list instead of restating a subset of it.

apps/web/src/store/appStore.ts
rsolmano: Applied in 6730a14 (declaration + action + doc comment on `modelsFresh`, and `store/SPEC.md` now names `dropModelsFreshness` as the third writer).

rsolmano:
Thanks — this was the sharpest of the non-blocking three, and all three are addressed in 6730a14.

1. **`reconcileModel`'s `models[0]`** — taken. It now returns `"unavailable"` (a verdict, not a replacement) and the dialog asks `model.default`, through the *same* `applyHostDefault` the preselect already used — so the host's `pinned ?? available[0]` policy exists in one place and the effort comes back consistent with the model it names, exactly as effort already defers to `model.clampThinking`. Asked at most once per opening (reset in the open-reset effect), so a registry race can't spin the effect; anything still wrong surfaces as `create()`'s error.
2. **The `models.length > 0` guard** — restored, on the read: the activation effect drops authority unconditionally but only fetches when the shared list is empty. Tab switches no longer trigger a host `runtime.refresh()`.
3. **`new-workspace.spec.ts`** — the spec now installs a `MutationObserver` on the refresh row before the click and asserts the recorded states *contain* `"true"`; a plain `toHaveAttribute("true")` would race the PI_OFFLINE round trip.

Gates: lint + typecheck + `check:deps`/`check:seams` green, 361 unit tests (2 new), `bun run e2e` 115/115 (one `history-search` flake on the first run, green in isolation).

packages/server/src/agent/piRuntime.ts
rsolmano: Confirmed and fixed in f9998cd — and the finding is broader than reported, so thank you.

`getAvailable()` in 0.82.1 (`model-runtime.js:211`):

```js
async getAvailable(providerId) {
  if (providerId) { … }
  await this.refreshAvailability();   // availabilityRefresh ?? queue(NEW pass)
  return this.snapshot.available;
}
```

So it not only awaits a *pending* availability pass, it **starts one** when none is pending. That means the same unbounded wait was sitting on `model.list` — whose documented contract is to answer without touching the network — as well as `model.default`, `resolveWireModel` and `oneshot.pickModel`, not just on the post-deadline read in `refreshAvailableModels`.

What landed:

1. **One settled read.** `settledAvailableModels(runtime)` → pi's `getAvailableSnapshot()`, used by every models read in `agent/`. That snapshot is pi's own result of the fan-out, written at `create()` (which awaits a refresh), after every `refresh()`, and on login/logout — so it is what the last *settled* pass concluded, and picker / default / ref-resolution now share one universe by construction. (`auth/providerStatus` deliberately keeps `getAvailable()`: there, awaiting the auth check *is* the answer.)
2. **Completion on the wire.** `refreshCatalogs` resolves `CatalogRefreshOutcome { completed }`, and `model.refresh` now answers `RefreshedModels { models, complete }` — v19 is new in this PR, so its note is amended rather than bumped. `finishModelsRefresh` sets `modelsFresh` **only** when `complete`, so a timed-out pass gives the picker a list to render and nothing to conclude from — which closes the "marks it fresh and reconciles against catalog state that never settled" path. `PI_OFFLINE` reports `completed: true` (nothing fetchable ⇒ the registry as it stands *is* settled).
3. **The wire-level test you asked for.** `agentSessionManager.test.ts`: `runtime.getAvailable` **and** `runtime.refresh` both never settle → `listAvailableModels()` still answers with the faux models, and `refreshAvailableModels(true)` resolves at the 15s ceiling with `complete: false`. Plus a store test that an unsettled list installs without granting authority.

Specs updated with it: `agent/SPEC.md` (the single settled read, and what `complete` means), `contracts/SPEC.md`, `store/SPEC.md`, `chat/SPEC.md`, `panels/SPEC.md`. Gates: lint + typecheck + `check:deps`/`check:seams`, 361 server + 309 web unit tests, `bun run e2e` 115/115.

packages/server/src/agent/SPEC.md
```diff
@@ -32,19 +32,43 @@ answer-injection path, and the **restart repair** that keeps re-opened transcrip
...
+    every picker's refresh row spinning. A timed-out caller serves the registry as it stands (reporting
+    `completed: false`) while single-flight keeps tracking the unbounded pass (so it cannot start a second concurrent refresh); failures `console.warn` + swallowed, never the picker's problem; **`PI_OFFLINE`**
+    (pi's env convention) disables it — resolving as a *completed* pass, since with nothing fetchable the
+    registry as it stands is the settled answer; the e2e webServer env and the manager's unit suite set it for
     hermeticity. The **provider-credential surface** over this runtime —
     `provider.status` + in-app login — lives in the sibling `auth` module (which consumes `getPiRuntime`),
     **not** here.
+
+    Every models **read** goes through **`settledAvailableModels(runtime)`** — pi's
+    `getAvailableSnapshot()`, **never `getAvailable()`**: that one awaits `refreshAvailability()`, which
```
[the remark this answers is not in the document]
rsolmano: Right, and worth fixing precisely because it's the security-sensitive sentence — a half-migrated spec is how a follow-up reintroduces the thing we just removed. Fixed in 726eea5: the inbound re-resolution now reads "via `resolveWireModel` against **`settledAvailableModels`** (the one settled read above — the picker's exact universe, so a ref the picker offered always resolves)". Swept the rest of the docs too: the only remaining `getAvailable()` mention in any spec is the one that exists to say *never* use it.

danyaberezun (approved):

PR #139, opened 2026-07-27
 Split the turn divider into specs vs changes, each linking the panel that owns it

OLavrik:
## Why                                                                                                                                                                                                         
                                                                                                                                                                                                                 
  A round that wrote only a spec said "1 file changed" and linked to Changes — which is git-derived and                                                                                                          
  can't show it: task-specs live in the gitignored `.thinkrail/context/`. The click landed on an empty view.                                                                                                     
                                                                                                                                                                                                                 
  ## What                                                                                                                                                                                                        
                                                                                                                                                                                                                 
  - Two chips instead of one: **specs** and **files changed**, each linking the panel that owns the artifact                                                                                                     
    ("is it a spec?" = membership in the same spec graph the Specs panel renders).                                                                                                                               
  - They act as a switch: one list open at a time, re-click clears, expanding reveals the owning view.                                                                                                           
  - Several artifacts expand inline in the chat — the set belongs to that round, not to the panels' *now*.                                                                                                       
  - Reuse pass: four copies of "read per workspace, re-read on fs tick" → `useWorkspaceRead`; path/array                                                                                                         
    primitives → `lib`; 12 destructure-deletes → `omitKey`. Net −4 lint suppressions.                                                                                                                            
                                                                                                                                                                                                                 
  ## Review notes                                                                                                                                                                                                
                                                                                                                                                                                                                 
  - `matchesWorktreePath` bug the new e2e caught: the suffix rule applied to relative paths too, so every                                                                                                        
    `<module>/SPEC.md` matched the root `SPEC.md`. Now absolute-only — also fixes a wrong-row highlight in Changes.                                                                                              
  - The `spec.graph` read moved to `RightPanel`: owned by the tab body it stopped tracking the worktree while                                                                                                    
    the user sat on Changes, silently undoing the split. Pinned by an `@agent` spec that fails without the fix.

### DIFF
--- a/apps/web/src/components/ErrorBoundary.tsx
+++ b/apps/web/src/components/ErrorBoundary.tsx
@@ -40,7 +40,7 @@ export class ErrorBoundary extends Component<Props, State> {
 	}
 
 	override componentDidUpdate(prev: Props): void {
-		if (this.state.error && !keysEqual(prev.resetKeys, this.props.resetKeys)) {
+		if (this.state.error && !shallowEqualArrays(prev.resetKeys, this.props.resetKeys)) {
 			this.reset();
 		}
 	}
@@ -62,9 +62,6 @@ export class ErrorBoundary extends Component<Props, State> {
 		);
 	}
 }
-
-/** Shallow (`Object.is`) equality of two `resetKeys` arrays — a caught error clears only when this returns false. */
-export const keysEqual = shallowEqualArrays;
 
 /** Themed, self-contained fallback — token utilities only, so it wears any theme. */
 function PanelErrorFallback({
--- a/apps/web/src/components/ErrorBoundary.tsx
+++ b/apps/web/src/components/ErrorBoundary.tsx
@@ -1,5 +1,6 @@
 import { AlertTriangle, RefreshCw, RotateCcw } from "lucide-react";
 import { Component, type ErrorInfo, type ReactNode } from "react";
+import { shallowEqualArrays } from "../lib";
 
 // Our single boundary primitive: contains a panel's render/lazy-import crash to that region instead of unmounting the root (bare gray `--bg-dark`); a rejected lazy `import()` (stale Vite chunk → 504) throws through Suspense into here, and we steer that case to a reload.
 
@@ -62,15 +63,8 @@ export class ErrorBoundary extends Component<Props, State> {
 	}
 }
 
-/** Shallow (`Object.is`) equality of two `resetKeys` arrays — a caught error clears only when this returns false. Pure so it's unit-testable. */
-export function keysEqual(
-	a: readonly unknown[] | undefined,
-	b: readonly unknown[] | undefined,
-): boolean {
-	if (a === b) return true;
-	if (!a || !b || a.length !== b.length) return false;
-	return a.every((value, i) => Object.is(value, b[i]));
-}
+/** Shallow (`Object.is`) equality of two `resetKeys` arrays — a caught error clears only when this returns false. */
+export const keysEqual = shallowEqualArrays;
 
 /** Themed, self-contained fallback — token utilities only, so it wears any theme. */
 function PanelErrorFallback({
--- a/apps/web/src/lib/utils.ts
+++ b/apps/web/src/lib/utils.ts
@@ -12,12 +12,14 @@ export function isMarkdownPath(path: string): boolean {
 }
 
 /**
- * Path separators as one form (`/`). Every path the app compares or displays arrives from a pi tool call or
- * the host, either of which may use the platform's separator — so normalizing is the first step of any path
- * predicate here, and it lives once for all of them.
+ * One canonical form for a path: `/` separators and no leading `./`. Every path the app compares or displays
+ * arrives from a pi tool call or the host, either of which may use the platform's separator and may prefix a
+ * relative path with `./` — so normalizing is the first step of any path predicate here, and it lives once
+ * for all of them. The `./` strip matters to *comparison*, not just display: without it a reported
+ * `./src/foo.ts` matches neither the entry `src/foo.ts` nor any spec-graph node.
  */
 export function normalizePath(path: string): string {
-	return path.replaceAll("\\", "/");
+	return path.replaceAll("\\", "/").replace(/^\.\/+/, "");
 }
 
 /** Posix or Windows absolute path — the two forms a tool call's `path` argument can arrive in. */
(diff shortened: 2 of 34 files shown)


rsolmano (commented):
Reviewed the full diff; gates are green on a clean worktree of the branch (`check:deps` + `lint` + `typecheck` + `test`), and the suppression audit comes back **-4** rather than +anything. The shape is good: the partition is derived in a pure deriver, the flip is one store field, the three worktree reads share one hook, and every SPEC claim I spot-checked matches the code.

Three findings worth fixing before merge, left as line comments:
1. `matchesWorktreePath` silently drops `./`-prefixed relative reports (a behavior regression vs the old suffix rule).
2. The spec deep-link fallback can open a **duplicate tab identity** for a file that is already open under its relative path.
3. The `shallowEqualArrays` extraction left `keysEqual` as a bare alias plus a duplicated equality test matrix.

Nits (not blocking, happy to file separately): `rightTabRequest` is never consumed, so re-activating a workspace re-fires the last flip and moves the user's tab; `useWorkspaceRead` subscribes to the whole store per instance (`subscribeWithSelector` would make the tick check free) and writes `latest.current` during render; `store/selectors.ts` imports `@/lib` while its sibling `appStore.ts` imports `../lib`; `ArtifactGroup.testid` doubles as the selection key and the expanded list has no `aria-controls` tie to its chip; and all five new e2e specs are `@agent`, so the fast gate covers none of the new UI — the switch spec in particular asserts exact chip text after a 4-write round, which is flake-prone.

apps/web/src/lib/utils.ts
```diff
@@ -11,6 +11,35 @@ export function isMarkdownPath(path: string): boolean {
 	return /\.(md|markdown)$/i.test(path);
 }
 
+/**
+ * Path separators as one form (`/`). Every path the app compares or displays arrives from a pi tool call or
+ * the host, either of which may use the platform's separator — so normalizing is the first step of any path
+ * predicate here, and it lives once for all of them.
+ */
+export function normalizePath(path: string): string {
+	return path.replaceAll("\\", "/");
```
rsolmano: **Behavior regression: `./`-prefixed relative paths stop matching.**

`matchesWorktreePath` now gates the suffix rule on `isAbsolutePath` (correct — that's the `<module>/SPEC.md` vs root `SPEC.md` fix), so a *relative* report has to be exactly equal. But the old rule in `ChangesPanel` was `c.path === want || want.endsWith("/" + c.path)`, which **did** absorb a `./` prefix: `"./src/foo.ts".endsWith("/src/foo.ts")` is `true`. After this change `./src/foo.ts` matches nothing — no Changes highlight, and it classifies as non-spec.

The prefix is collateral damage of the anchoring fix, and it belongs in the primitive that every predicate starts from:

```suggestion
	return path.replaceAll("\\", "/").replace(/^\.\//, "");
```

Worth a case in `utils.test.ts` (`normalizePath("./src/foo.ts") === "src/foo.ts"`) and in `selectors.test.ts` (`matchesWorktreePath("./src/foo.ts", "src/foo.ts") === true`), since that's exactly the pair the anchoring test now pins in the other direction.

apps/web/src/panels/SpecsPanel.tsx
```diff
@@ -9,66 +8,54 @@ import {
...
-			cancelled = true;
-		};
-	}, [workspaceId, fsTick, refreshToken]);
+		if (specRequest?.workspaceId !== workspaceId) return;
+		const want = specRequest.path;
+		const store = useAppStore.getState();
+		const known = store.specsByWorkspace[workspaceId]?.find((node) =>
+			matchesWorktreePath(want, node.path),
+		);
+		void openFileInTab(workspaceId, known?.path ?? want);
```
rsolmano: **The `?? want` fallback can open a second tab for an already-open file.**

The documented reason for the fallback is "a spec created seconds ago" — i.e. the path *as pi reported it*, which may be **absolute**. That read succeeds server-side (`resolveInWorktree` does `resolve(root, path)`, and an in-worktree absolute path resolves to itself), so the failure is invisible: `openFileInTab` derives the tab id as `` `${workspaceId}:${path}` ``, so the tab lands as `w1:/Users/…/wt/SPEC.md` while every other entry point (tree, markdown links, Changes) uses `w1:SPEC.md`. Result: two tabs with the same `name` (basename) for one file, each stale-checked independently against the fs tick — and the de-dupe/focus guarantee `openFileInTab`'s own doc comment makes ("`path` is worktree-relative … both get identical de-dupe/focus behavior") is broken by this call site rather than by the helper.

Cleanest fix is at the choke point, so no future caller can reintroduce it — normalize an in-worktree absolute path to relative inside `openFileInTab` (it already reads the workspace out of the store for the tab lookup, so the `worktreePath` is at hand). If you'd rather keep `openFileInTab` strict, resolve it here instead, e.g. strip the active workspace's `worktreePath` prefix from `want` before the `??`, and leave a comment that the relative form is the tab identity.

apps/web/src/components/ErrorBoundary.tsx
```diff
@@ -62,15 +63,8 @@ export class ErrorBoundary extends Component<Props, State> {
...
-export function keysEqual(
-	a: readonly unknown[] | undefined,
-	b: readonly unknown[] | undefined,
-): boolean {
-	if (a === b) return true;
-	if (!a || !b || a.length !== b.length) return false;
-	return a.every((value, i) => Object.is(value, b[i]));
-}
+/** Shallow (`Object.is`) equality of two `resetKeys` arrays — a caught error clears only when this returns false. */
+export const keysEqual = shallowEqualArrays;
```
rsolmano: **Finish the extraction instead of aliasing it.**

`keysEqual` is now a pure re-export of `shallowEqualArrays` whose only remaining job is the old name — an indirection that makes the boundary look like it owns a comparison it no longer does. And the sweep left the tests duplicated: `ErrorBoundary.test.ts` still asserts the whole matrix (`NaN`, `-0`, `undefined` on either side, length mismatch) that `lib/utils.test.ts` now owns verbatim.

Suggested collapse:

```suggestion
```

…then use the primitive directly at the one call site (line 43):

```ts
if (this.state.error && !shallowEqualArrays(prev.resetKeys, this.props.resetKeys)) {
```

…and trim `ErrorBoundary.test.ts` to the intent it uniquely pins — that `resetKeys` gates auto-recovery (equal keys keep the error, a changed key clears it) — dropping the `NaN`/`-0`/`undefined` cases now covered in `lib`. `components/SPEC.md`'s public surface line should then lose `keysEqual` too.

rsolmano (approved):

Issue #11, opened 2026-07-06
We need some nice way to show draft text for approval, or changs


rsolmano:
Could be applied for the all changes in general. 
We need to show user changes and the ability to somehow interact with them

PR #141, opened 2026-07-27
fix(dialog): the Windows folder picker opens in front, and a failed picker says why

danyaberezun:
## The report

On Windows, **Open project** does nothing: no folder dialog, no error.

## No merge broke this

The first suspicion was a regression in one of today's merges (two stable releases were cut today — v0.0.9 at 13:08, v0.0.10 at 21:09). It isn't. Across `v0.0.8 → v0.0.10` the entire picker path is untouched:

```
git diff v0.0.8 v0.0.10 -- packages/server/src/dialog \
    apps/web/src/panels/{AddProjectMenu,useOpenProject}.tsx apps/web/src/transport
 apps/web/src/panels/useOpenProject.tsx | 3 +++      <- a comment, nothing else
```

- `packages/server/src/dialog/**` last changed **2026-06-30** (#11); Windows binaries have shipped since **2026-07-07** (#73).
- `handlers.ts`'s `"dialog.selectDirectory"` line: unchanged. `AddProjectMenu`: unchanged since #49.
- Checked out `origin/main` in a scratch worktree and ran its own e2e for that exact click path (`projects.spec.ts` + `welcome.spec.ts`, 12 specs driving `add-project-menu` → `menu-open-project`): green.

So the picker has been broken on Windows since it was written. The one *related* change today is #135, which removed the "Open project" card from the Welcome screen when a project is shown — that moves **where** you click, not **how** the dialog is invoked.

## Two causes, both original

**1. Z-order.** We spawn the picker from a background process, which Windows forbids from taking the foreground — so an *unowned* `FolderBrowserDialog` opens **behind** the browser the user is looking at. Indistinguishable from a dead button. It's now owned by an invisible top-most form (`$d.ShowDialog($owner)`), which puts it in front.

**2. Silence.** Every failure read as "user cancelled": a non-zero exit was treated as a cancel, and `pickAndOpen` swallowed the rest with `// Cancelled / unavailable — nothing to do`. But PowerShell exits **0** on cancel — a non-zero exit there is a real failure (a blocked `Add-Type` under ConstrainedLanguage, no host on PATH). `Picker` now declares `nonZeroExit` per picker (`osascript` -128, `zenity`, `kdialog` genuinely do cancel that way); a failure, or no runnable candidate at all, **throws** with the reason, and the hook surfaces it in the existing `NoticeDialog`. The picker is the only way to add a project, so a silent `null` is a dead button.

Also `-Sta` (a WinForms dialog needs a single-threaded apartment — the default for `powershell.exe`, *not* for `pwsh`) and a `pwsh.exe` fallback, matching the hosts and order `apps/cli/src/powershell.ts` already uses.

## Not verified on Windows

No Windows box here, so the z-order fix is a best diagnosis of a symptom I can't reproduce. That's exactly why the error surfacing is in scope: if it still fails, the next report names a cause instead of nothing. Sanity check on a Windows terminal — if this shows a dialog but the app doesn't, the cause is elsewhere and the new error text will say so:

```
powershell.exe -NoProfile -Sta -Command "Add-Type -AssemblyName System.Windows.Forms; $d = New-Object System.Windows.Forms.FolderBrowserDialog; if ($d.ShowDialog() -eq 'OK') { $d.SelectedPath }"
```

## Test plan

`check:deps`, `check:seams`, `lint`, `typecheck`, `bun run test` (349), `bun run e2e` (115) — all green. Unit tests pin the two hosts, `-Sta`, the owner form, and the cancel-vs-failure split per platform.

UI-visible surface is the failure path only (the existing open-error notice, now also carrying picker failures) — happy to attach a forced-failure screenshot if useful.

### DIFF
--- a/apps/web/src/panels/SPEC.md
+++ b/apps/web/src/panels/SPEC.md
@@ -45,7 +45,9 @@ arrangement (so the mobile shell is an additive layer, not a rewrite).
   `WelcomePanel`, so the flow is identical in the rail and the Welcome screen): `project.open`, and on
   failure `project.inspect` → either offers to bootstrap the folder into a repo — a modal **`ConfirmDialog`**
   (confirm → `project.init`) — when it's `initable`, or surfaces the error in a **`NoticeDialog`** — so a
-  non-git folder is never a silent no-op. Both are modals on `components/ui/dialog` (the init offer has no
+  non-git folder is never a silent no-op — and neither is a host that couldn't *show* a folder dialog (that
+  throws; the notice carries the reason, and the request runs on a raised `timeoutMs` since the picker waits
+  on a human). Both are modals on `components/ui/dialog` (the init offer has no
   on-screen anchor, unlike the Remove popover); `NoticeDialog` is a single-button info modal for failures
   with no yes/no follow-up. The hook returns a `dialogs` node each consumer renders. **Selecting a
   project** (clicking its row — the chevron expands/collapses separately) **deselects any active
--- a/apps/web/src/panels/useOpenProject.tsx
+++ b/apps/web/src/panels/useOpenProject.tsx
@@ -64,12 +64,16 @@ export function useOpenProject(onOpened: (project: Project) => void | Promise<vo
 
 	/** Ask the host for a directory via its native picker, then open it. */
 	const pickAndOpen = async () => {
+		let path: string | null;
 		try {
-			const { path } = await getTransport().request("dialog.selectDirectory", {});
-			if (path) await openProject(path);
-		} catch {
-			// Cancelled / unavailable — nothing to do.
+			({ path } = await getTransport().request("dialog.selectDirectory", {}));
+		} catch (err) {
+			// A cancel is a null path; a throw means the host couldn't *show* a dialog — surface it, or the
+			// only way to add a project reads as a dead button.
+			setOpenError(errorText(err, "Couldn't open the folder picker on the host."));
+			return;
 		}
+		if (path) await openProject(path);
 	};
 
 	const dialogs = (
--- a/apps/web/src/panels/useOpenProject.tsx
+++ b/apps/web/src/panels/useOpenProject.tsx
@@ -5,6 +5,11 @@ import { errorText, getTransport } from "../transport";
 import { ConfirmDialog } from "./ConfirmDialog";
 import { NoticeDialog } from "./NoticeDialog";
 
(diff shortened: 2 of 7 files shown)


danyaberezun:
Thanks — the Windows verification is the part I couldn't do, and both follow-ons were real. Fixed in `1c649fd`.

**1. The 60s request timeout defeated the picker.** Confirmed diagnosis: a folder dialog is held open by a human, so `request` rejected at 62s *while the dialog was still on screen*, and the deleted pending entry meant the host's eventual reply was dropped. Worse after this PR than before, since the old `catch {}` swallowed it and now it surfaces as a misleading error.

`request` now takes a per-call `timeoutMs`. The unused third positional `sessionId` became a `RequestOptions` bag — no call site passed it, so no call sites changed:

```ts
request(method, params, { timeoutMs: PICK_TIMEOUT_MS })
```

The picker raises it to **30 min**: generous for a human, finite so a dead host still releases the request (pending entries aren't rejected on socket close, so "no timeout" would leak one). I stopped short of the `startLogin`/`resolveLogin` split — it's the right shape for this, but a bigger change than this fix warrants; happy to do it as a follow-up if you'd rather not carry a long-blocking request at all.

**2. The failure message could come out empty.** Right on both counts — `split("\n")` always yields at least `""`, so the `??` was dead and a killed picker produced the bare `"The folder picker failed: "`, plus a trailing `\r` from PowerShell's CRLF. Now `err.replaceAll("\r", "").trim().split("\n")[0] || \`exit ${code}\``.

**3. Missing coverage on the throw paths** — which is exactly why #2 slipped through. Both message builders are extracted (`pickerFailure`, `noPickerMessage`) and unit-tested: CRLF stripping, empty stderr → `exit 137`, whitespace-only stderr, and the Linux "install zenity" vs. named-platform branch.

**4. The `-Sta` comment was wrong** for pwsh 7+ (STA is the default there). Comment corrected to say both hosts already default to STA on Windows and the flag only pins the requirement at the spawn site. I kept the flag rather than dropping it, since your 4/4 verification ran with it — I didn't want to ship a command line different from the one proven on the actual machine.

**5. `panels/SPEC.md`** now records the picker-failure notice path; `transport/SPEC.md` records the `timeoutMs` override and why a human-answered request must raise it.

Gates on this branch: `check:deps`, `check:seams`, `lint`, `typecheck`, 349 unit tests, **115/115 e2e** — green here on Linux, so the two you couldn't run (the `core.autocrlf` formatter noise and Playwright launch) are covered.

PR #143, opened 2026-07-28
feat(models): order the model picker newest-first and adopt pi's enabledModels allowlist

### LINKED ISSUE #195: model list: filter, prioritise, and sort

### LINKED ISSUE #206: allow lists and permissions

rsolmano:
## Why

The model picker showed pi's catalog verbatim — on my machine that's **60 models**, with
`gpt-4` / `gpt-4-turbo` / `gpt-4o-2024-05-13` / `o1` leading the OpenAI group and **every Anthropic model
listed twice** (the alias plus its dated twin, e.g. `claude-opus-4-5` + `claude-opus-4-5-20251101`).

pi's SDK offers no help here, and I checked before inventing anything: `ModelRuntime.getAvailable()`
filters only by provider, and `Model` (pi-ai 0.82.1) carries **no release date and no deprecation flag** —
pi's own sorting is alphabetical (`--list-models`) or "current model first" (TUI). But pi *does* own an
allowlist for exactly this complaint: `settings.json` → **`enabledModels`** (ordered glob patterns, same
format as `--models`, editable from pi's `/models`), which this app never read.

So: order host-side, and adopt pi's allowlist rather than inventing a "what looks outdated" policy.

## What changed

**Host** — new `packages/server/src/agent/modelCatalog.ts`, which now owns *everything model-shaped*
(the wire projection, inbound ref re-resolution, pi's thinking clamp, default-model resolution) so
`agentSessionManager` is session lifecycle only; the edge between them is one-way.

- **Order**: provider asc, then `(version digits desc, dated snapshot directly under its alias, id asc)`.
  Version digits are the only recency signal that exists in pi's data.
- **`enabled`**: pi's `enabledModels`, resolved with pi's own `resolveModelScopeWithDiagnostics` (globs,
  dedupe, alias-over-dated) — so a stored pattern means the same thing in the app and in the pi CLI.
  **Fails open**: unmatched patterns warn and are swallowed; an all-stale list reads as "no allowlist"
  rather than an empty picker.
- **`collapsed`**: a dated snapshot whose alias is also listed — computed **only when no allowlist is
  set** (an explicit list is the curation; we never second-guess it).
- Writes go to pi's **global** settings; "everything enabled" is stored as an absent setting (pi's own
  semantics). `model.default` now skips a disabled model.

**Wire (v19)** — `model.list` → `ModelCatalogEntry[]` (`WireModel` stays clean: it also rides
`session.create` / `SessionSummary`, where allowlist membership is meaningless), new `model.setEnabled`,
new `model.catalogChanged` broadcast (converge-on-broadcast, like `settings.changed`). `modelRef()` is the
one `"provider/id"` formatter both ends share.

**Web** — the picker renders the two host-decided tiers (everyday list + a `Show all N models` row), and a
**search query crosses both tiers**, so a legacy or pinned model is never unreachable. New
**Settings → Models** section curates the list (per-model toggles, "Enable all"), saving through the
broadcast — no dirty state, no Save button. The tier split lives in a store selector, memoized on the
catalog's identity so it can't churn renders.

## Result on a real catalog

Dumped from the running UI (e2e):

- **Everyday tier** starts `claude-fable-5, claude-opus-5, claude-sonnet-5, claude-opus-4-8, …` then
  `gpt-5.6-luna … gpt-5.1, gpt-5, gpt-4.1, gpt-4, gpt-4o, o4-mini, o3, o1` — legacy families sink on their own.
- **Hidden behind "Show all"**: exactly the 7 dated duplicates (`claude-*-2025…` ×4, `gpt-4o-2024-*` ×3).

## Tradeoff, taken knowingly (pi parity)

A save persists the client's list verbatim, so a signed-out provider's refs drop out and a
newly-authenticated provider's models start disabled while a list is set — the same behavior as pi's own
`/models`. Merge semantics were considered and rejected: a preserved glob (`claude-*`) would silently
re-enable the very model the user just switched off. Recorded in the agent SPEC.

## Verification

- `packages/server/src/agent/modelCatalog.test.ts` — 18 cases: the exact ordering rule (incl. date-suffix
  stripping), the collapse rule, glob/stale allowlist resolution, `setEnabledModels` writing/clearing real
  pi settings in a temp agent dir + broadcasting, `model.default` preferring an enabled model.
- New `selectors.test.ts` cases (tier split + selector reference stability) and a `formatContextWindow` case.
- `e2e/models.spec.ts` — picker tiers, "Show all", search across tiers, and the Settings → Models
  round-trip; guarded on a resolved model, since the no-agent suite deliberately runs without provider auth.
- Full gates: `check:deps`, `check:seams`, `lint`, `typecheck`, `bun run test` (361 server), `bun run e2e`
  (**117 passed**).
- `resetState` now strips `enabledModels` from the isolated agent dir — it shares `settings.json` with the
  `@agent` suite's pinned default model, which `model.default` resolves *through* the allowlist.

## Specs

Spec-first, then code: `packages/server/src/agent/SPEC.md`, `packages/contracts/SPEC.md`,
`apps/web/src/{chat,panels,store,lib}/SPEC.md`.

## Screenshots

Captured from the real UI and waiting on disk at `/tmp/thinkrail-shots/` — drag them into this PR to
attach (the CLI can't upload binaries):

| shot | what it shows |
|---|---|
| `before/picker.png` | today's picker: `gpt-4`, `gpt-4-turbo`, `gpt-4o-2024-05-13` … up top, Claude models doubled |
| `after/picker.png` | the everyday tier: newest first per provider, duplicates gone, `Show all 53 models` row at the bottom |
| `after/picker-show-all.png` | expanded — the dated snapshots under a "More models" heading |
| `after/settings-models.png` | the new Settings → Models manager |

### DIFF
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -14,6 +14,7 @@ import { Popover, PopoverAnchor, PopoverTrigger } from "@/components/ui/popover"
 import {
 	EMPTY_RUNTIME,
 	SettingsSection,
+	selectPickerTiers,
 	selectSkillsStale,
 	selectWorkspaceById,
 	specPathMatcher,
@@ -124,7 +125,8 @@ export default function ChatView({
 	// This tab's runtime — zustand only re-renders when *this* session's slice ref changes, so a background
 	// chat streaming into its own runtime never re-renders the foreground one.
 	const runtime = useAppStore((s) => s.sessions[sessionId]) ?? EMPTY_RUNTIME;
-	const models = useAppStore((s) => s.models);
+	const catalogSize = useAppStore((s) => s.modelCatalog.length);
+	const { primary: models, extra: extraModels } = useAppStore(selectPickerTiers);
 	// This chat's owning project (workspaces are keyed by project) — for the Skills manager's trust ops
 	// and the "project" / "all" history-search scopes.
 	const projectId = useAppStore(
@@ -245,14 +247,15 @@ export default function ChatView({
 	const chatLocationRequest = useAppStore((s) => s.chatLocationRequest);
 	const [flashRowId, setFlashRowId] = useState<string | null>(null);
 
-	// Models are global to the host — fetch once, then every chat's picker shares them.
+	// The catalog is global to the host — fetch once, then every chat's picker shares it (later writes
+	// arrive on the `model.catalogChanged` broadcast).
 	useEffect(() => {
-		if (models.length > 0) return;
+		if (catalogSize > 0) return;
 		getTransport()
 			.request("model.list", {})
-			.then((m) => useAppStore.getState().setModels(m))
+			.then((catalog) => useAppStore.getState().setModelCatalog(catalog))
 			.catch(() => {});
-	}, [models.length]);
+	}, [catalogSize]);
 
 	// The skill catalog is per-session; load it when the chat opens.
 	useEffect(() => {
@@ -680,6 +683,7 @@ export default function ChatView({
 							mentionCandidates={mentionCandidates}
 							recentPrompts={recentPrompts}
 							models={models}
+							extraModels={extraModels}
 							currentModel={currentModel}
 							thinkingLevel={thinkingLevel}
 							onMentionQuery={onMentionQuery}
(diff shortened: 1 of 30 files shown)


danyaberezun:
should close #195 and #206

PR #162, opened 2026-07-30
feat(analytics): message_sent event on every user send

rsolmano:
## What

Adds a fifth member to the closed analytics event union: **`message_sent`**, fired once per user-authored send — a new chat's first message and any message in an existing chat alike (`chat_started` remains the new-chat signal).

```ts
| { name: "message_sent"; params: { mode: SendMode } }   // SendMode = "prompt" | "steer" | "follow_up"
```

## Design decisions

- **`mode` only.** Closed vocabulary mirroring pi's three send methods. No text, no length, no image count, and no identity params — model preference is already answered by `chat_started`, which also kept `agent` free of a new session-model accessor.
- **Accepted sends only.** `trackSend(mode)` sits *after* `await ackSend(...)` in `session.prompt` / `session.steer` / `session.followUp`, so a rejected send (bad model, missing key, malformed send) never counts.
- **`session.answerQuestion` untracked** — a tool reply, not a user message.
- **Call site stays in `host`.** Per `submodule-server-analytics`, feature modules remain analytics-free.

## Privacy

The machine-checked pin is `analytics.test.ts`: `EVENT_SAMPLES` is exhaustive over the union (a new variant without a sample is a compile error) and `EXPECTED_KEYS` asserts the exact outgoing property set — here `app_version`, `channel`, `os`, `arch`, `mode`.

## Specs updated

- `packages/server/src/analytics/SPEC.md` — event list + a new "Engagement" bullet (vocabulary, accepted-only firing, why no identity params).
- `packages/server/src/host/SPEC.md` — the new `track()` call site.
- `README.md` — Analytics & Privacy section.

## Verification

`check:deps` · `check:seams` · `lint` · `typecheck` · `bun run test` (365 pass) · `bun run e2e` (124 pass) · `spec_validate` clean. No suppressions added.

Not covered by a test: the accepted-path fire needs a live pi session (`@agent` territory); the negative path holds structurally from `track` sitting after `await ackSend`. Happy to add an `@agent` e2e assertion if you'd like it pinned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,8 @@ stored in `~/.thinkrail/installation.json`; it never leaves the host except as t
 `distinct_id` on events. Events additionally carry only low-cardinality, non-personal metadata: app
 version, release channel (`stable`/`nightly`), OS (`macos`/`linux`/`windows`), architecture
 (`x64`/`arm64`), and — on chat/login events — the model/provider name **only if it is a pi built-in**
-(anything user-configured is reported as `custom`). Events are sent **personless** (no person
+(anything user-configured is reported as `custom`). Message activity is counted as a bare send event
+carrying only *how* it was sent (`prompt`/`steer`/`follow_up`) — never the message itself. Events are sent **personless** (no person
 profiles are ever built) and with **GeoIP lookup disabled**.
 
 **Never collected:** file paths or names, prompts, code, chat transcripts, API keys, token counts,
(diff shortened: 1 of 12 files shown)


danyaberezun (dismissed):

rsolmano:
Addressed in 3943ebb — thanks, the finding was correct.

`useChatTodos.nudgeAgent` does ride `session.prompt` / `session.followUp`, so the original placement counted the app's own control prompts as user messages (add a TODO item, send nothing, still get a `message_sent`).

Fix: the wire already marks that traffic with `TODO_NUDGE_PREFIX`, but it was being read with a bare `startsWith` in two places (host history index, web hydrate). Centralised it as contracts' **`isControlMessage(text)`**, migrated both existing readings to it, and `trackSend(mode, text)` now returns early through the same predicate — so the three consumers of "what counts as internal traffic" can't drift.

`session.answerQuestion` remains untracked (tool reply). Specs updated: `contracts/SPEC.md` (public surface + why one shared reading), `analytics/SPEC.md` (user-authored-only), `host/SPEC.md` (the call-site condition). Re-ran: `check:deps`, `check:seams`, `lint`, `typecheck`, `bun run test` (365 pass), `bun run e2e` (124 pass).

Issue #12, opened 2026-07-06
AskUserQuestion has "review" but it is reachable only manually using mice


rsolmano:
Either remove it at all, or don't skip and ask user to submit explicitly

rsolmano:
Fixed by #66 (merged 2026-07-13, commit 1274a28): for multi-question questionnaires the final question's **Next** now routes to the **Review & submit** step instead of skipping past it, and **Submit** is only exposed from review — so the review step is part of the keyboard/sequential flow, no longer reachable only by mouse. Single-question cards keep their direct Submit.

PR #185, opened 2026-08-09
fix(terminal): key shells to their tab, with the host owning the mapping

danyaberezun:
Stacked on #181 — review that first.

## The bug

Leaving and re-entering a workspace faster than one round trip spawned a **second shell** and left the first running with nothing pointing at it, for the life of the host.

The client held the only copy of the tab→shell mapping and removed the entry *before* asking the host whether that shell was still alive. A remount inside that window found nothing and created another.

Reproduced by holding `terminal.alive` for 3s and re-entering twice:

```
distinct pty ids:   [ '9899f3dd-…', 'f7c7740f-…' ]   ← two shells, one tab
original id frames: 105 → 109 → 112                   ← orphan still running
CHECK=$TR_SURVIVOR → "CHECK="                         ← shell state gone
```

The deeper cause is that the client inferred user intent at unmount time — *"was my tab closed, or did the panel go away?"* — from a store read. That guess is the root of the whole bug family, not just the race.

## The fix

Rather than guard the window, remove the state that can go stale.

| | |
|---|---|
| **`terminal.attach`** | Idempotent get-or-create keyed by `(workspaceId, tabKey)`. Replaces `terminal.create` + `terminal.alive`. Mount is one call; unmount does nothing. |
| **Host owns the tab list** | `terminal.list`. The client can never hold the only record of a running shell. |
| **Owner-scoped shells** | Not per-page — matching `history`/`todos`/`templates`, which already assume a single-owner host. Shells survive a reload, a closed browser, a different browser. |
| **Exclusive attach** | A PTY has one size, so a takeover notifies the displaced client (`terminal.detached`) rather than silently reflowing it. Mirroring stays additive. |
| **Replay on attach** | A bounded server-side recorder repaints the screen, so a surviving shell no longer comes back behind a blank pane. |
| **Revive** | A host restart cannot preserve shells; tabs come back with fresh ones showing the recording. |
| **Reference-based GC** | No tab → no shell, plus host lifetime. The abandoned-client reap is **gone** — it was what made a reload destructive. |
| **Confirm on close** | Closing a tab is the only client-driven kill and refuses a busy shell until confirmed — checked and killed in one synchronous pass, so nothing started in between dies unannounced. |

Atomicity comes from the attach handler being **synchronous** on Bun's single event loop — the per-session mutex for free. Pinned by test and SPEC, because an `await` added later would silently reintroduce double-spawn.

**Deleted:** `detachedPtyByClientId`, `terminal.alive`, `isTerminalAlive`, `closeClientTerminals`, the `tabSurvives` inference, and both stale-closure branches.

## Design check

Cross-checked against VS Code (`attachToProcess`, keyed `(workspaceId, instanceId)`; recorder replay via `persistentSessionScrollback`; reconnection vs revive), tmux (`new -A -s`), Jupyter/terminado (`GET/POST /api/terminals`), and Zed's daemon RFC. Theia is the cautionary case — frontend-held id against an in-memory backend registry produces `Couldn't attach - can't find terminal with id`, which our server-side get-or-create makes unrepresentable.

tmux as the persistence layer was considered and **rejected** (recorded in `architecture.md` #12): unassumable on Windows, competing tab model, env-propagation breakage, polling-based capture — and we have already accepted no crash isolation.

No idle culling, deliberately: Jupyter's `cull_inactive_timeout` measures last PTY I/O, not whether a process runs, and kills quiet long-running commands.

## Verification

- `bun run e2e` — **170 passed**, including new specs for the rapid re-entry race, reload survival, second-client takeover, and both close-confirm paths.
- `bun run test` — 473 unit tests; new: `outputRecorder`, `shellBusy`, `terminalManager` (attach idempotency incl. concurrent, takeover, busy-close, revive).
- `check:deps`, `check:seams`, `lint`, `typecheck` green. No suppressions added.

Protocol v27.

settings-terminal-dark
busy-close-dark
detached-light
detached-dark
after-replay-dark

### DIFF
--- a/apps/web/src/panels/SPEC.md
+++ b/apps/web/src/panels/SPEC.md
@@ -78,12 +78,12 @@ arrangement (so the mobile shell is an additive layer, not a rewrite).
   with no yes/no follow-up. The hook returns a `dialogs` node each consumer renders. **Selecting a
   project** (clicking its row — the chevron expands/collapses separately) **deselects any active
   workspace**, so the shell returns to that project's Welcome — a deliberate "project home" gesture; the
-  workspace's tabs survive in the store, so re-selecting it restores its view. **Terminals need more than the
-  store to survive that round trip**, since the whole workspace surface — `TerminalsPanel` included —
-  unmounts: each instance therefore **detaches** its PTY instead of closing it
-  (`detachedPtyByClientId` in `TerminalInstance`) and the next mount re-adopts the same shell, so a
-  long-running process is never silently killed by a project-home gesture. Only a *closed tab* kills its PTY.
-  The painted scrollback does not survive (a remount is a fresh xterm buffer); the process does. Also
+  workspace's tabs survive in the store, so re-selecting it restores its view. That round trip unmounts the
+  whole workspace surface, `TerminalsPanel` included — but **terminals keep no client-side state to lose**:
+  the host owns the tab list and keys shells by `(workspaceId, tabKey)`, so mounting is one idempotent
+  `terminal.attach` and unmounting does nothing at all. Attach returns the recorded output to repaint, so the
+  screen comes back too. **Only `terminal.close`, from the user closing a tab, kills a PTY** — and it refuses
+  a shell with child processes until a `ConfirmDialog` is accepted. Also
   `FileTree`, `SpecsPanel`, `RightPanel`,
   `ChangesPanel` (the changed files under a header that says **what** is being diffed — the
   **`ChangesScopeMenu`** scope pill + the shared **`BranchPicker`** target-branch pill — plus the
--- a/apps/web/src/panels/SPEC.md
+++ b/apps/web/src/panels/SPEC.md
@@ -81,7 +81,10 @@ arrangement (so the mobile shell is an additive layer, not a rewrite).
   workspace's tabs survive in the store, so re-selecting it restores its view. That round trip unmounts the
   whole workspace surface, `TerminalsPanel` included — but **terminals keep no client-side state to lose**:
   the host owns the tab list and keys shells by `(workspaceId, tabKey)`, so mounting is one idempotent
-  `terminal.attach` and unmounting does nothing at all. Attach returns the recorded output to repaint, so the
+  `terminal.attach` and unmounting does nothing at all. **Exactly one `TerminalInstance` is mounted app-wide —
+  the tab on screen.** Mounting *is* attaching and attachment is exclusive, so an instance per tab would claim
+  terminals this client is not showing and take them from the client that opened them; switching tabs
+  re-attaches and repaints from the host's recording. Attach returns the recorded output to repaint, so the
   screen comes back too. **Only `terminal.close`, from the user closing a tab, kills a PTY** — and it refuses
   a shell with child processes until a `ConfirmDialog` is accepted. Also
   `FileTree`, `SpecsPanel`, `RightPanel`,
--- a/apps/web/src/panels/SettingsDialog.tsx
+++ b/apps/web/src/panels/SettingsDialog.tsx
@@ -6,6 +6,7 @@ import {
 	Palette,
 	ShieldCheck,
 	SlidersHorizontal,
+	SquareTerminal,
 } from "lucide-react";
 import { Dialog, DialogContent, DialogHeader, DialogTitle } from "@/components/ui/dialog";
 import { cn } from "@/lib";
@@ -15,12 +16,14 @@ import { GithubSettings } from "./GithubSettings";
 import { PrivacySettings } from "./PrivacySettings";
 import { ProvidersSettings } from "./ProvidersSettings";
 import { TemplatesSettings } from "./TemplatesSettings";
+import { TerminalSettings } from "./TerminalSettings";
 
 /** The live settings sections, in nav order. */
 const SECTIONS: { id: SettingsSection; label: string; icon: LucideIcon }[] = [
 	{ id: SettingsSection.Providers, label: "Providers", icon: KeyRound },
 	{ id: SettingsSection.Github, label: "GitHub", icon: GitBranch },
 	{ id: SettingsSection.Appearance, label: "Appearance", icon: Palette },
+	{ id: SettingsSection.Terminal, label: "Terminal", icon: SquareTerminal },
 	{ id: SettingsSection.Templates, label: "Templates", icon: LayoutTemplate },
 	{ id: SettingsSection.Privacy, label: "Privacy", icon: ShieldCheck },
 ];
@@ -96,6 +99,8 @@ export function SettingsDialog() {
 							<ProvidersSettings />
 						) : section === SettingsSection.Github ? (
 							<GithubSettings />
+						) : section === SettingsSection.Terminal ? (
+							<TerminalSettings />
 						) : section === SettingsSection.Templates ? (
 							<TemplatesSettings />
 						) : section === SettingsSection.Privacy ? (
--- a/apps/web/src/panels/TerminalInstance.tsx
+++ b/apps/web/src/panels/TerminalInstance.tsx
@@ -120,7 +120,6 @@ interface Props {
 	/** This tab's durable identity. Half of `(workspaceId, tabKey)`, which is what the host keys shells on. */
 	tabKey: string;
 	workspaceId: string;
-	visible: boolean;
 	/**
 	 * Run once, on the shell this tab was opened for (e.g. "Open in Vim"), then spent.
 	 *
@@ -140,7 +139,7 @@ interface Props {
  * Attach returns the recorded output to repaint, so a remount shows the screen it left behind rather than an
  * empty buffer over a live process.
  */
-export default function TerminalInstance({ tabKey, workspaceId, visible, initialCommand }: Props) {
+export default function TerminalInstance({ tabKey, workspaceId, initialCommand }: Props) {
 	const hostRef = useRef<HTMLDivElement>(null);
 	const termRef = useRef<XTerm | null>(null);
 	const serverIdRef = useRef<string | null>(null);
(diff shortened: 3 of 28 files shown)


danyaberezun:
### Follow-ups pushed (dfad412)

**Windows child-process detection.** `hasChildProcesses` now has a third probe — CIM `Win32_Process` filtered by parent pid — so the confirm-before-closing-a-running-shell guard is no longer absent on Windows. Two deliberate details: the pid rides an **env var**, never the command string (interpolating into PowerShell's parser is what `apps/cli/src/powershell.ts` exists to avoid), and the script **prints a count** rather than signalling by exit code, because PowerShell exits 0 whether or not the query matched — an exit-code answer could not distinguish "no children" from "the query failed", and we would report *not busy* on a host where the check actually works. I cannot execute this path from Linux; what is unit-tested is the part that does not need Windows (both properties above).

**Configurable replay budget.** `AppConfig.terminalReplayKb` — default 64, max 1024, `0` disables replay entirely — with a new **Settings → Terminal** section. Clamped by its consumer on read against `TERMINAL_REPLAY_KB`, so a hand-edited `config.json` cannot size an unbounded per-terminal buffer; the settings bag persists what it is given.

**One more bug fixed in review.** `initialCommand` was gated on the host's `created` flag alone — but `created` is also true when a tab's *previous* shell exited and the attach spawned a replacement, so an "Open in Vim" tab would have reopened vim on every revisit. It is now spent from the store the first time it runs, matching its documented "sent once, never replayed".

### Verification

- `bun run e2e` — **170 passed** (dev host)
- `bun run e2e:binary` — **166 passed** against the compiled single-file binary, the gate for the regression class that only exists inside the artifact. Relevant here because this change adds a persistence file (`terminals.json`) and boot-time work (`reviveTerminalSessions`).
- `bun run test` — unit suites green, including new recorder / `shellBusy` / `terminalManager` coverage
- `check:deps`, `check:seams`, `lint`, `typecheck` green; no suppressions added

Screenshots of the three new surfaces (attached-elsewhere overlay in both themes, busy-close confirm, Settings → Terminal) are captured and will be attached separately.

danyaberezun:
### Review findings addressed (8562ec6)

All four confirmed against the code — every one was real. Three were the same shape: authority moved to the host and enforcement did not follow it there.

**1. Only the attached client may drive a terminal.** `terminal.write`/`terminal.resize` lost their caller check when ownership moved from per-client to owner-scoped. A displaced client keeps a valid PTY id, and the transport replays unresolved frames across a reconnect — so a keystroke queued before a takeover could execute in whoever holds the tab now. Both now take the caller and no-op unless it matches `attachedClient`.

On the "reject vs auto-reattach" choice I went with **reject**, following the art: `tmux attach -d` and VS Code's `requestDetachInstance`/`acceptDetachInstanceReply` both make reclaiming an explicit gesture, and neither lets a displaced client silently steal a terminal back by typing. "Take it back" stays the way back.

**2. Tab-list changes are broadcast.** The host owned the list but never published changes. New `terminal.tabs` channel: an idempotent per-workspace snapshot on open/close/archive, folded by every client — the workspace-trio pattern from architecture #9. A shell's *contents* stay addressed to the one attached client; only which terminals *exist* is shared.

**3. The store no longer resurrects deleted tabs.** `pending` was a guess ("missing from the host list ⇒ attach in flight"). Now tracked explicitly: `TerminalTab.attachPending`, set by `addTerminal` and cleared by `settleTerminalAttach` when the attach lands. Only genuinely-pending tabs are exempt from an authoritative list.

**4. Alt-screen detection is a stream** (non-blocking, fixed anyway). Incomplete escape sequences now carry across PTY reads and the stream is split at each transition, so a `?1049h` split across reads or an enter+exit sharing one is handled. New tests cover both.

That last one surfaced a bug of its own: the switch sequence was being recorded, so a replay would have emitted `?1049h` and flipped the fresh terminal to the alt screen — showing nothing at all. Mode sequences are now excluded from the recorded body entirely; `snapshot`'s preamble already re-emits the ones worth restoring.

Also fixed while here: closing a tab the host does not know about (an attach that never landed, or one another client already closed) left an undismissable row, because the client only dropped it on `closed: true`.

### Verification

- `bun run e2e` — **171 passed**, including a new spec for cross-client convergence: B opens a tab → A sees it; B closes it → A converges, rather than keeping a dead instance that accepts input.
- Unit: new coverage for displaced-client rejection (write is dropped, then works again after an explicit reclaim), the broadcast firing on membership change but *not* on every re-attach, split-sequence and same-chunk alt-screen transitions.
- `check:deps`, `check:seams`, `lint`, `typecheck` green; no suppressions.

Specs updated: `terminal/SPEC.md` (attached-client restriction, the broadcast exception to addressing, stream-based recorder rules) and `store/SPEC.md` (`attachPending`).

danyaberezun:
### Round 2 addressed (8ff2407)

Both confirmed. Threads resolved.

**Blocking — a stale attach could outrank a takeover.** Real, and the reachable path is worse than plain reordering. `terminal.detached` is fire-and-forget, so a client that was mid-reconnect during a takeover never receives it at all — and its *replayed* attach comes back from the request-replay cache with the original success. The pane then goes ready over a PTY that now rejects everything it sends: apparently live, entirely inert.

Fixed on both sides, since either alone leaves a hole:
- the client tracks an **attach generation**, bumped on every `terminal.detached`, and drops an attach response whose generation has moved — so a later detach always wins;
- the host **re-announces `terminal.detached`** to any caller that tries to write or resize a terminal it is not attached to, so a lost notice heals on the first keystroke rather than never.

**Non-blocking — a dead tab lost its screen.** Also real. `pty.onExit` deleted the entry holding the recorder while the tab lived on, so the next attach opened a blank pane and `persistTerminalSessions` had nothing to save — losing exactly the output that would say why the shell died. The final screen now moves into the tab's pending-replay slot before the recorder goes, so it survives both a remount and a host restart. `revivedOutput` is renamed `pendingReplay`, since it is no longer only about revival.

One note on the first finding: the pure push-before-response ordering you describe is not itself reachable on a healthy socket — the host's reply to A is written while handling A's frame, strictly before B's frame is read, and WebSocket preserves per-connection order. The defect is real via the replay path instead, and the generation guard covers both orderings regardless.

**Verification:** `bun run e2e` 171 passed; `terminalManager.test.ts` now 19, adding displaced-client re-announcement (and that the attached client is *not* told), last-screen retention across a remount, and across a host restart. Lint/typecheck green — lint is in fact what caught that my first attempt at the generation guard never wired up its increment, leaving it inert.

danyaberezun:
### Round 3 addressed (06bc60f)

Confirmed and fixed. Thread resolved.

**Blocking — the persisted tab list was stale for the whole run.** Real, and sharper than it might read: the host has **no crash isolation by design** (a fatal agent or provider fault takes it down — `architecture.md`), so an exit without `stop()` is an ordinary path here, not an edge case. Close a tab, take a provider fault, and the next boot reloaded the old file, brought the tab back, and mounting it spawned a shell for a tab the user had closed — exactly the no-tab/no-shell invariant this PR introduced. Newly opened tabs vanished by the same mechanism.

Membership is now written wherever it changes — new-tab attach, close, workspace archive — folded into the same `membershipChanged` step that already fired the broadcast, so the clients and the file learn at the same moment and cannot disagree. Output snapshots stay best-effort, as you suggested: they ride along when they exist, and `stop()` still captures a full set before killing the shells. An unclean exit therefore gives back the right *tabs* with blank screens, which is the honest degradation.

The three new tests deliberately never call `persistTerminalSessions()` — skipping it is precisely what models the crash.

**Verification:** `bun run e2e` 171 passed; `terminalManager.test.ts` now 22; lint/typecheck/deps/seams green.

danyaberezun:
### Round 4 addressed (84b34f4)

Confirmed — and it was a hole I opened by adding the reclaim path without giving it its own buffer.

The prebind buffer is single-use (inert once bound), but "Take it back" reused the one created at mount, while detaching had already cleared the pty id. So anything arriving before a reclaim's response was dropped twice over. Your reconnect scenario is the reachable one: the host resumes a held `terminal.exit` in its `open` handler *before* it re-serves the replayed attach from the request-replay cache, so the exit hit an inert buffer and a null id and vanished — then the cached success marked the pane ready over a shell that was already dead.

Each attempt now takes a fresh buffer and binds only its own, so a stale attempt can never retire a newer one's, and the generation guard and the buffer now agree on which attempt owns the response.

**Covered as asked.** New e2e: B holds the tab and arms the shell to die, A reclaims with its attach response held past the death, and the pane must come back `data-exited="true"`. I checked it is a real regression test — reverting just `TerminalInstance.tsx` makes it fail, and restoring it makes it pass.

**Verification:** `bun run e2e` **172 passed**. One unrelated flake in `history-search.spec.ts` appeared on the first full run; it passes in isolation both with and without this change, and the next full run was clean — so it is flake in that spec, not a regression here. Lint/typecheck/deps/seams green; no suppressions anywhere in the branch.

danyaberezun:
### Round 5 addressed (f649569)

Confirmed, and a good catch — it is an interaction between two things this PR did, neither wrong alone.

Rendering an instance per tab is left over from when that was how xterm buffers survived a tab switch. Once the tab list became *shared*, it turned harmful exactly as you describe: a `terminal.tabs` broadcast made every other browser render a hidden instance, and since mounting **is** attaching and attachment is exclusive, that hidden instance snatched the terminal out from under whoever had just opened it.

Fixed the way you suggested: **exactly one instance is mounted app-wide — the tab on screen.** Switching tabs re-attaches and repaints from the host's recording, which is what the recording exists for; a background shell keeps running and keeps being recorded. The flatten/visibility helpers (`allTerminalTabs`, `isTerminalVisible`) and the `visible` prop only a mount-everything panel needed are deleted, so the panel is smaller than before.

You were also right that the convergence test could not see this — it only counted rows. It now asserts the **opener keeps its terminal** and can still type in it after the other browser has rendered the new row.

One clarification, since it looks similar but is not the same thing: in that test client A *is* detached on the first terminal, because B entered the workspace looking at it. That is exclusive attach working as designed, not a steal — the bug was specifically about claiming tabs a client is **not** showing. I noted it in the test so the next reader does not mistake one for the other.

**Verification:** `bun run e2e` **172 passed**; unit suites green; lint/typecheck/deps/seams clean.

danyaberezun:
@jetbrains-air please re-review — rebased onto the updated #181 (6e631ac). Your approval here was auto-dismissed with "The merge-base changed after approval" when #181 was force-pushed, so this is the same content on a new base rather than new work.

The only conflicts were in the two files #181 also touched: the protocol changelog in `wsProtocol.ts` (this branch's attach redesign renumbered v27 → **v28**, since #181 now holds v26 and v27) and the replay-cache paragraph in `packages/server/src/host/SPEC.md`, where this branch's "replay-namespace retention timer" wording was kept and #181's now-detailed bound description replaced the stale "count-and-serialized-byte-bounded" prefix. No behavioural change on either side.

Gates on the rebased head: lint, typecheck, `check:deps`, `check:seams`, 509 server + 420 web unit tests, 174/174 no-agent e2e.

PR #210, opened 2026-08-10
Chat: image attachments that can't brick a session, file-chip rendering, and a retry that doesn't duplicate the reply

OLavrik:
Summary                                                                                                                                                                                                        
                                                                                                                                                                                                                 
  Four independent fixes to the chat surface, all in apps/web/src/chat plus one server-side guard. The headline is the image-attachment fix: a single oversized pasted image used to permanently break a chat —  
  pi re-sends the whole history every turn, so once Anthropic's cap rejected that image, every subsequent turn 400'd and the conversation was unrecoverable.                                                     
                                                                                                                                                                                                                 
  Oversized images — two lines of defense                                                                                                                                                                        
                                                                                                                                                                                                                 
  Anthropic caps an image side at 8000px, dropping to 2000px once a single request carries more than 20 images. The composer was sending pasted images raw, and pi's own resizer is deliberately off server-side 
  (images.autoResize:false — the photon/WASM bundling problem).                                                                                                                                                  
                                                                                                                                                                                                                 
  Client — apps/web/src/chat/imageAttachment.ts (new). Pasted and dropped images are decoded in the browser and downscaled to a 1568px long edge (Claude's own standard-tier edge) before becoming ImageContent. 
  Within-bounds images pass through byte-identical; an undecodable file falls back to raw. The pending chip shows mime · W×H.                                                                                    
                                                                                                                                                                                                                 
  Server — packages/server/src/agent/imageGuard.ts (new). An inline pi extension on the context event (fired before every LLM call, live sessions included) sniffs each image block's dimensions straight from   
  the base64 header bytes — PNG/JPEG/GIF/WebP, no codec, never strips what it can't sniff — and replaces any block over the cap with a text note carrying the W×H and a re-attach hint.                          
                                                                                                                                                                                                                 
  Sessions are append-only ("pi owns state") and the host has no image codec, so the guard transforms the outgoing context only — the session file and the visible transcript stay untouched. That's what        
  un-bricks the already-poisoned chats: they recover on their very next message. The count-aware cap also self-heals the read-tool case — a raw 3000px read is legal while the context holds ≤20 images and      
  degrades to a note (not a brick) once the session crosses 21.   
  
  Attachments render as file chips                                                                                                                                                                               
                                                                                                                                                                                                                 
  UserTurn shows image blocks as compact "attached file" chips above the text rather than inline previews; clicking opens the image in a dialog (the diagram-fullscreen pattern). The chip label is the picked   
  file's name, carried on the echo turn as attachmentNames — UI-side only, since pi's ImageContent has no filename field — index-aligned with the image blocks. A hydrated turn has no names and falls back to   
  mime-type labels.                                                                                                                                                                                              
                                                                                                                                                                                                                 
  New ChatAttachment type (name + content) threads the filename from composer to echo turn without touching the wire.                                                                                            
                                                                                                                                                                                                                 
  No horizontal scroll in the transcript                                                                                                                                                                         
                                                                                                                                                                                                                 
  Long unbroken tokens in the user bubble (URLs, pasted paths) pushed it wider than the container. break-words on the bubble and overflow-x-hidden on the Virtuoso scroller — wide content scrolls inside its own
  code/diff block, never the whole chat.                                                                                                                                                                         
                                                                                                                                                                                                                 
  Auto-retry no longer duplicates the reply                                                                                                                                                                      
                                                                                                                                                                                                                 
  pi's _prepareRetry removes the failed attempt's assistant message from the transcript before re-running the turn, and the retried run re-streams the reply as a brand-new message. The reducer only showed a   
  countdown on auto_retry_start and kept the failed attempt's turn, so the client rendered the reply twice (frozen failed partial + retried copy) while pi's transcript — and any reloaded client hydrating it — 
  held one. The reducer now mirrors pi's slice: drop the trailing assistant turn on auto_retry_start, skipping client-only retry-countdown turns, conservative like pi's own last-message guard.                 
                                                                                                                                                                                                                 
  Testing                                                                                                                                                                                                        
                                                                                                                                                                                                                 
  - Unit: imageAttachment.test.ts (downscale math, passthrough, fallback), imageGuard.test.ts (hand-built header bytes per format, both caps, the count threshold), rows.test.ts (attachment names on derived    
  rows), appStore.test.ts — including a red-first reducer test replaying pi's exact retry event sequence (message_end error → agent_end willRetry → auto_retry_start → retried stream ⇒ exactly one assistant    
  turn) plus the error-before-message_start edge.                                                                                                                                                                
  - E2E: e2e/composer-images.spec.ts (no-agent — attach, chip metadata, downscale assertions). A true e2e repro of the retry bug needs a mid-turn provider network failure the real-provider-only agent suite    
  can't induce; instead e2e/hydrate-midstream.live.spec.ts (@agent) pins the neighbouring invariant that falsified the first hypothesis — reloading mid-stream hydrates exactly one copy of the streaming        
  message.                                                                                                                                                                                                       
  - Gates (all green on the rebased branch): check:deps, check:seams, lint, typecheck, test (570 pass / 0 fail).                                                                                                 
                                                                                                                                                                                                                 
  Specs updated: apps/web/src/chat/SPEC.md, apps/web/src/store/SPEC.md, packages/server/src/agent/SPEC.md.

### DIFF
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -660,7 +660,8 @@ export default function ChatView({
 							context={listContext}
 							components={CHAT_LIST_COMPONENTS}
 							// `overflow-x-hidden` on the scroller: the chat only ever scrolls vertically — wide
-							// content (code, diffs) scrolls inside its own block, never the whole transcript.
+							// content (code, diffs, GFM tables — see Markdown's `Table` wrapper) scrolls inside its
+							// own block, never the whole transcript.
 							className="min-h-0 flex-1 overflow-x-hidden"
 							// Any chat opens at the latest message (a fresh mount would otherwise land mid-transcript);
 							// the jump-to-message deep link overrides post-mount with its centered scrollToIndex.
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -1,6 +1,5 @@
 import type {
 	AskUserQuestionResult,
-	ImageContent,
 	PromptHit,
 	SlashCommandInfo,
 	TemplateInfo,
@@ -45,7 +44,7 @@ import { stripFrontmatter } from "./templateText";
 import { useModelCatalog } from "./useModelCatalog";
 import "./tools/register"; // side-effect: register the built-in pi tool renderers (bash/read/edit/write)
 import { ChatTurnView } from "./turns";
-import type { ChatTurn } from "./types";
+import type { ChatAttachment, ChatTurn } from "./types";
 import { useChatScroll } from "./useChatScroll";
 import { useChatTodos } from "./useChatTodos";
 import { useHistorySearch } from "./useHistorySearch";
@@ -362,9 +361,10 @@ export default function ChatView({
 			.catch(() => {});
 	};
 
-	const onSubmit = (text: string, images: ImageContent[], behavior: SubmitBehavior) => {
-		if (text || images.length > 0)
-			useAppStore.getState().appendUserMessage(sessionId, text, images);
+	const onSubmit = (text: string, attachments: ChatAttachment[], behavior: SubmitBehavior) => {
+		if (text || attachments.length > 0)
+			useAppStore.getState().appendUserMessage(sessionId, text, attachments);
+		const images = attachments.map((a) => a.content);
 		const params = { sessionId, text, ...(images.length > 0 ? { images } : {}) };
 		const method =
 			behavior === "steer"
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -659,7 +659,9 @@ export default function ChatView({
 							data={rows}
 							context={listContext}
 							components={CHAT_LIST_COMPONENTS}
-							className="min-h-0 flex-1"
+							// `overflow-x-hidden` on the scroller: the chat only ever scrolls vertically — wide
+							// content (code, diffs) scrolls inside its own block, never the whole transcript.
+							className="min-h-0 flex-1 overflow-x-hidden"
 							// Any chat opens at the latest message (a fresh mount would otherwise land mid-transcript);
 							// the jump-to-message deep link overrides post-mount with its centered scrollToIndex.
 							initialTopMostItemIndex={{ index: Math.max(rows.length - 1, 0), align: "end" }}
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -363,7 +363,8 @@ export default function ChatView({
 	};
 
 	const onSubmit = (text: string, images: ImageContent[], behavior: SubmitBehavior) => {
-		if (text) useAppStore.getState().appendUserMessage(sessionId, text);
+		if (text || images.length > 0)
+			useAppStore.getState().appendUserMessage(sessionId, text, images);
 		const params = { sessionId, text, ...(images.length > 0 ? { images } : {}) };
 		const method =
 			behavior === "steer"
--- a/apps/web/src/chat/Composer.tsx
+++ b/apps/web/src/chat/Composer.tsx
@@ -53,6 +53,13 @@ interface PendingImage extends AttachedImage {
 	name: string;
 }
 
+/** A refused pick, rendered as a dismissible error chip: the file's name plus why it couldn't attach. */
+interface AttachError {
+	id: string;
+	name: string;
+	reason: string;
+}
+
 /** The token (non-whitespace run) ending at the caret — drives `@`-mention completion. */
 function activeToken(value: string, caret: number): { token: string; start: number } {
 	const match = /(\S+)$/.exec(value.slice(0, caret));
@@ -248,8 +255,9 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 	const [pendingImages, setPendingImages] = useState(0);
 	// Files that could NOT be attached (undecodable + provider-unsupported type, or over the request-wide
 	// image budget) — surfaced as dismissible error chips in the attachment strip; silently dropping a
-	// pick would read as a successful attach.
-	const [attachErrors, setAttachErrors] = useState<{ id: string; text: string }[]>([]);
+	// pick would read as a successful attach. Name and reason stay separate so the chip can truncate the
+	// (user-controlled) filename while keeping the reason visible.
+	const [attachErrors, setAttachErrors] = useState<AttachError[]>([]);
 	const [mentionActiveIndex, setMentionActiveIndex] = useState(0);
 	const [mentionDismissed, setMentionDismissed] = useState(false);
 	// The plain `↑`-recall session: `null` when inactive; otherwise an index into `recentPrompts` (0 =
@@ -462,22 +470,16 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 			// over-budget turn is rejected forever (the host guard heals history, but never sending is cheaper).
 			let used = imagesRef.current.reduce((sum, p) => sum + p.content.data.length, 0);
 			const additions: PendingImage[] = [];
-			const errors: { id: string; text: string }[] = [];
+			const errors: AttachError[] = [];
 			settled.forEach((result, i) => {
 				const name = picked[i]?.name || "image";
 				if (result.status !== "fulfilled" || result.value === null) {
-					errors.push({
-						id: crypto.randomUUID(),
-						text: `Couldn't attach ${name} — unsupported image format`,
-					});
+					errors.push({ id: crypto.randomUUID(), name, reason: "unsupported image format" });
 					return;
 				}
 				const size = result.value.content.data.length;
 				if (used + size > REQUEST_IMAGE_BASE64_BUDGET) {
-					errors.push({
-						id: crypto.randomUUID(),
-						text: `Couldn't attach ${name} — message image limit reached`,
-					});
+					errors.push({ id: crypto.randomUUID(), name, reason: "message image limit reached" });
 					return;
 				}
 				used += size;
@@ -725,7 +727,11 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 							data-testid="composer-image-error"
 							tone="error"
 							icon={false}
-							label={err.text}
+							// The filename truncates, the reason never does — the reason is the only thing the
+							// user can act on, and a phone has no tooltip to fall back to.
+							title={`Couldn't attach ${err.name} — ${err.reason}`}
+							label={`Couldn't attach ${err.name}`}
+							meta={`— ${err.reason}`}
 							trailing={
 								<button
 									type="button"
@@ -745,12 +751,9 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 							data-width={img.width}
 							data-height={img.height}
 							data-mime={img.content.mimeType}
-							label={
-								<>
-									{img.name}
-									{img.width && img.height ? ` · ${img.width}×${img.height}` : null}
-								</>
-							}
+							title={img.name}
+							label={img.name}
+							meta={img.width && img.height ? ` · ${img.width}×${img.height}` : undefined}
 							trailing={
 								<button
 									type="button"
--- a/apps/web/src/chat/Composer.tsx
+++ b/apps/web/src/chat/Composer.tsx
@@ -12,6 +12,7 @@ import {
 	useRef,
 	useState,
 } from "react";
+import { FileChip } from "./FileChip";
 import { type AttachedImage, fileToAttachedImage } from "./imageAttachment";
 import { ModelSelector } from "./ModelSelector";
 import {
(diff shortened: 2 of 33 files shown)


apps/web/src/chat/imageAttachment.ts
```diff
@@ -0,0 +1,96 @@
...
+export async function fileToAttachedImage(file: File): Promise<AttachedImage> {
+	let bitmap: ImageBitmap;
+	try {
+		bitmap = await createImageBitmap(file);
+	} catch {
+		return { content: await fileToRawContent(file) };
+	}
+	try {
+		const { width, height } = fitWithin(bitmap.width, bitmap.height, MAX_ATTACHMENT_EDGE);
+		if (width === bitmap.width && height === bitmap.height) {
```
danyaberezun: **Blocking** — there is no byte ceiling here or in the guard, so the session-bricking case this PR targets survives.

Measured: a 1280×960 / 8 s / 20 fps screen-capture GIF (9.3 MB) through the real attach path in Chromium — `createImageBitmap` reports 1280×960, long edge ≤ 1568, so this raw branch returns it byte-identical: **12.42 MB of base64** sent as `image/gif`. `imageGuard` compares only `width`/`height`, so it does not strip it either. The provider limit is 5 MB per image — pi says so at `utils/image-resize-core.js:3`.

The half that matters is the guard: a `data.length` predicate alongside the dimension one, so a poisoned session still self-heals. Here, a JPEG quality ladder for within-bounds files that are already large.

apps/web/src/chat/ChatView.tsx
```diff
@@ -629,7 +630,9 @@ export default function ChatView({
 							data={rows}
 							context={listContext}
 							components={CHAT_LIST_COMPONENTS}
-							className="min-h-0 flex-1"
+							// `overflow-x-hidden` on the scroller: the chat only ever scrolls vertically — wide
+							// content (code, diffs) scrolls inside its own block, never the whole transcript.
+							className="min-h-0 flex-1 overflow-x-hidden"
```
danyaberezun, line 665: **Blocking** — this clips wide content instead of making it scroll in its own block. Code blocks and tool cards do have their own scroller; GFM tables do not (`Markdown.tsx:15-16` gives a table only `border-collapse` plus cell borders, and the generated typography adds only fonts).

Measured in Chromium, 14-column table in a 400px scroller, real wheel gesture: pre-PR `overflow-x` computed to `auto` (the inline `overflowY: auto` forces it) and scrolled to `scrollLeft` 324; with `hidden` it stays at 0 — while `scrollWidth` is 724 against `clientWidth` 400 in both cases. The overflow is still there, just unreachable.

Fix: wrap `table` in an `overflow-x-auto` div via `Markdown`'s `components`. (`[&_table]:block` also stops the bleed but destroys table layout.)

apps/web/src/chat/imageAttachment.ts
```diff
@@ -0,0 +1,96 @@
...
+	if (longEdge <= maxEdge) return { width, height };
+	const scale = maxEdge / longEdge;
+	return {
+		width: Math.max(1, Math.round(width * scale)),
+		height: Math.max(1, Math.round(height * scale)),
+	};
+}
+
+/** Mime types canvas.toDataURL can (re-)encode; anything else re-encodes as PNG when resized. */
+const CANVAS_ENCODABLE = new Set(["image/png", "image/jpeg", "image/webp"]);
```
danyaberezun, line 63: **Blocking** — this set only takes effect on the resize branch, so a within-bounds file keeps `file.type`, and the composer filter admits any `image/*`.

Measured: a 1200×900 BMP passes the filter, decodes, is within bounds → sent as `image/bmp`. pi converts everything outside png/jpeg/gif/webp to PNG (`utils/image-process.js:6-27`), and `imageDimensions` returns `undefined` for BMP so the guard cannot strip it. The same BMP *over* 1568px is handled correctly — only the small ones break.

Fix: re-encode whenever `file.type` is outside the four accepted types, resize needed or not.

apps/web/src/chat/Composer.tsx
```diff
@@ -426,10 +407,17 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 	const addFiles = async (files: File[]) => {
 		const picked = files.filter((f) => f.type.startsWith("image/"));
 		if (picked.length === 0) return;
-		const contents = await Promise.all(picked.map(fileToImageContent));
+		// Downscaled at attach time (≤1568px long edge) — an oversized image in history 400s every
+		// subsequent turn once the provider's many-image cap kicks in. See imageAttachment.ts.
+		const attached = await Promise.all(picked.map(fileToAttachedImage));
```
danyaberezun: **Blocking** — `setImages` runs only after this await, `addFiles` is invoked as `void addFiles(...)` (`:563`, `:570`), and neither `submitText` (`:323`) nor the send button's `disabled` (`:836`) knows about work in flight. Press Enter inside that window and the message goes without the image — and the chip then lands on the *next* message.

Measured pipeline cost: 30 ms (1600×1200) to 141 ms (4000×3000) on desktop headless Chromium, wider on the phone this client targets first. Paste-then-Enter is the normal gesture.

Fix: a pending counter in state, surfaced as a placeholder chip and honoured by `disabled`/`submitText`.

apps/web/src/chat/imageAttachment.ts
```diff
@@ -0,0 +1,96 @@
...
+ * loss); a resized one re-encodes in its own format when canvas supports it, else PNG. A file the
+ * browser can't decode falls back to the raw bytes — attaching must never fail here; the server-side
+ * guard still protects the session.
+ */
+export async function fileToAttachedImage(file: File): Promise<AttachedImage> {
+	let bitmap: ImageBitmap;
+	try {
+		bitmap = await createImageBitmap(file);
+	} catch {
+		return { content: await fileToRawContent(file) };
```
danyaberezun, line 105: This fallback can reject — `fileToRawContent` rejects on a `FileReader` error — which contradicts the docstring's "attaching must never fail here". Under the caller's `Promise.all` that discards every sibling which decoded fine, and `void addFiles(...)` has no `.catch`, so the user sees nothing and the page takes an unhandled rejection. `allSettled`, or resolve to `null`.

packages/server/src/agent/imageGuard.ts
```diff
@@ -0,0 +1,198 @@
...
+	const blocksOf = (m: AgentMessage): ContentBlock[] | undefined => {
+		const content = (m as { content?: unknown }).content;
+		return Array.isArray(content) ? (content as ContentBlock[]) : undefined;
+	};
+
+	// One bounded sniff per image block per pass, cached by block identity — the hook fires on every LLM
+	// call, so re-decoding per predicate (count / detect / note text) would multiply the work.
+	const sniffed = new Map<ContentBlock, Dimensions | undefined>();
+	let imageCount = 0;
+	for (const message of messages) {
```
danyaberezun, line 169: `imageCount` is fixed before anything is replaced, so the strip invalidates the count that chose the cap. Measured on the real function: 18 images at 500px plus 3 at 2500px → all three stripped, and the request ends up carrying 18. Dropping one would have taken it to 20, lifting the cap back to 8000px and leaving the other two legal. Low priority — strip largest-first only down to the threshold, then re-evaluate.

apps/web/src/chat/turns.tsx
```diff
@@ -94,18 +96,87 @@ export function ChatTurnView({
...
+function AttachmentChip({ label, img }: { label: string; img: ImageContent }) {
+	const [open, setOpen] = useState(false);
+	return (
+		<>
+			<button
+				type="button"
+				data-testid="chat-attachment-chip"
+				title="View image"
+				onClick={() => setOpen(true)}
+				className="flex items-center gap-xs whitespace-nowrap rounded-[var(--radius-sm)] border border-border-default bg-container-elevated-bg px-sm py-xs text-text-default tr-text-metadata transition-colors hover:bg-control-bg-hovered focus-visible:ring-2 focus-visible:ring-primary"
```
danyaberezun: These nine utilities and the `FileIcon` + label structure are repeated from `Composer.tsx:651` (this one adds `whitespace-nowrap` and interaction states). One `FileChip` taking a label plus an optional trailing slot would serve both and stop the composer and transcript chips drifting apart.

packages/server/src/agent/imageGuard.test.ts
```diff
@@ -0,0 +1,290 @@
...
+	});
+
+	test("reads GIF logical screen dimensions", () => {
+		expect(imageDimensions(gifBytes(2500, 900).toString("base64"))).toEqual({
+			width: 2500,
+			height: 900,
+		});
+	});
+
+	test("reads WebP VP8X canvas dimensions", () => {
```
danyaberezun, line 138: Only `VP8X` is covered — the lossy `VP8 ` and lossless `VP8L` branches (`imageGuard.ts:76-86`) have no test, and they are the densest bit arithmetic in the file. I checked both against the container specs and they are correct, which is exactly why a regression there would be silent.

e2e/hydrate-midstream.live.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { expect, test } from "@playwright/test";
+import { createWorkspaceViaDialog, openFixtureProject, worktreeRows } from "./fixtures/app";
+
+// Tagged @agent (see agent.live.spec.ts): a REAL pi agent. Pins the mid-stream hydration invariant:
+// a client that hydrates while a turn streams (reload, second tab, workspace switch) must end up with
+// exactly ONE copy of the streaming assistant message. This holds by construction today —
+// `session.getMessages` returns only COMMITTED messages (pi keeps the in-flight partial in
+// `streamingMessage`, pushing it to the transcript at message_end), and the reducer's adopt-by-mint
+// path builds the live turn once — but a pi bump or a hydration change could silently break either
+// half, and the symptom (a duplicated reply) is exactly the class of bug users report as "дубляж".
```
danyaberezun: Non-English word at the end of this line — the only non-ASCII text in the diff.

Separately: this spec covers an adjacent invariant that its own comment calls true by construction, at ~180 s of real provider tokens in the `@agent` suite. Worth a deliberate keep-or-drop rather than a default.

danyaberezun (changes requested):
Requesting changes: four verified defects, three of them in the size-guarding path this PR is about. Everything below was reproduced by running the real module or a real browser — details in each comment.

**Blocking**
1. No byte ceiling in either layer — an ordinary animated GIF reaches the provider at 12.42 MB against a 5 MB limit. `chat/imageAttachment.ts:82`
2. `overflow-x-hidden` makes a wide markdown table unreachable rather than tidy. `chat/ChatView.tsx:635`
3. A within-bounds BMP/AVIF/HEIC is sent with a media type the provider rejects. `chat/imageAttachment.ts:41`
4. A send during attach-time decoding silently drops the attachment. `chat/Composer.tsx:412`

**Nits, not blocking**
- `turns.tsx:142` — `alt={label}` reads out as "image/png" on the hydrated fallback; prefer `alt=""` plus an `aria-label` on the button.
- `chat/SPEC.md:310-314` — five added lines start at column 0 inside an indented bullet.
- `packages/contracts` has no test files, so `isRetriedAttempt` is covered only transitively. Five cases in a `domain.test.ts` would pin the shared contract where it lives.
- `domain.ts:487` — the forward scan treats `toolResult`/`custom` as transparent, wider than the docstring. No reachable case found, but `messages[index + 1]?.role === "assistant"` states exactly the guarantee `_prepareRetry` gives.
- `attachmentNames` couples two arrays by index across store → row → component. Correct today; `attachments: { name, content }[]` on the turn would make the pairing structural.

**Checked and cleared**, so these don't get re-raised: sniffer cost is negligible (20 sniffs of a 2 MB payload = 1.2 ms, so the per-pass cache is right) · a malformed JPEG returns `undefined`, not a bogus size · attachment names survive tab/workspace switches (`hydrateSession` bails on a live runtime) · the retry trim cannot fire on a `length` stop (`_isRetryableError` requires `stopReason === "error"`) · the image dialog is safe inside a virtualized row (Radix modal locks scroll) · the guard reaches reopened sessions · compaction is not a hole (it serializes to text) · the `context` hook is the right seam and the copy-on-write is real · `read`-tool images are in scope · history-index alignment holds · no suppressions in the diff.

All four fast gates pass once `node_modules` matches the lockfile. Browser suite not run.

danyaberezun:
**races:**
- `apps/web/src/chat/Composer.tsx:442` — `let used = images.reduce(...)` runs *after* `await Promise.allSettled(...)` on the closure-captured `images`, so two overlapping `addFiles` calls each seed the request budget from the same pre-batch snapshot. **Confirmed empirically** with 1000×1000 noise PNGs (4,588,588 B of base64 each, passthrough; budget 25,165,824 B): one paste of 6 → 5 chips + 1 error (in-batch accumulator correct); two pastes of 3 in the same tick → **6 chips, 0 errors = 27,531,528 B attached, 2.37 MB over budget, no refusal**. The comment above the check ("a chip removed during the decode window only makes the check conservative, never over-budget") covers removals but not concurrent additions. Fix: synchronous ref accumulator, or serialize `addFiles`.

**duplications:**
- `apps/web/src/chat/Composer.tsx:702` re-hand-rolls the chip skin from `apps/web/src/chat/FileChip.tsx:7` — `flex items-center gap-xs whitespace-nowrap rounded-[var(--radius-sm)] border … px-sm py-xs … tr-text-metadata` verbatim, only the border/bg/text tokens differ. `FileChip`'s own docstring says it exists as "a single source so the two can't drift apart"; a `tone` prop folds the error chip back in.
- `apps/web/src/chat/imageAttachment.ts:71` and `:83` — the data-URL→base64 split (`indexOf(",")` + conditional `slice`) is written twice; `fileToRawContent` can delegate to `dataUrlToContent`.
- `apps/web/src/chat/Composer.tsx:341`+`:921` — the send-blocked predicate exists twice (`if (pendingImages > 0) return; … if (!text && images.length === 0) return;` vs `disabled={pendingImages > 0 || (!value.trim() && images.length === 0)}`). Pre-existing shape, but this PR added the `pendingImages` term to both copies instead of deriving one `canSubmit`.
- `packages/server/src/agent/imageGuard.ts:188,191,236,239` — the MB formatting (`/(1024*1024)`, `.toFixed(1)`) repeats across two note builders.

rsolmano (dismissed):

rsolmano (dismissed):

rsolmano (approved):

PR #243, opened 2026-08-16
layout: add IDE-style workspace layouts

danyaberezun:
## What
- Adds recursive center splits, resizable/foldable side groups, tab drag/drop/search, presets, and terminals in center or side groups.
- Synchronizes durable workspace layouts across clients while keeping focus, drafts, and attention browser-local.
- Makes concurrent full-layout replacements conflict-safe instead of allowing a stale last writer to silently overwrite a newer layout.
- Integrates the durable TODO review map as a live plan page whose shared placement stores only the plan kind and session identity.

## How
- Adds protocol v40 exact-base replacement: `expectedRevision: null` is create-only and a number replaces only that accepted revision; the merged TODO review-map decoration remains the preceding v39 change.
- Compares the expected revision inside the server's per-workspace queue before validation or persistence; stale writes return a typed conflict carrying the current snapshot and never increment or broadcast.
- Chains client optimistic revisions, rolls back a conflicting mutation and its dependents without stale-document retry, settles lost responses from matching broadcasts, and prevents delayed conflicts from regressing newer accepted state.
- Keeps semantic layout mutations in `shell/layout`, and decomposes workbench orchestration into bounded `layoutSync`, `layoutIntents`, `chatReconciliation`, `terminalReconciliation`, and `legacySelection` modules.
- Makes primary-modifier tests Apple-aware, hardens parallel auto-rename/history-recall fixtures, and synchronizes post-drag/remote-history E2Es on observable readiness without blanket retries or weakened assertions.

## Screenshots
| Workbench | Side split target |
|---|---|
| Recursive workbench | Side split target |

| Tab search | Presets |
|---|---|
| Tab search | Layout presets |

## Verification
- 1,440 unit tests passed.
- 224/224 no-agent E2E tests passed across 8 isolated shards.
- Dependency pins, binary seams, lint, typecheck, spec validation (apart from two pre-existing dangling links), and diff/suppression audits passed.


.github/pr-assets/ide-workbench/layout-presets.png
rsolmano, line 1: Is it a good idea to store those files in git? (couldn't suggest better option for now though)

PR #251, opened 2026-08-17
Add start_new_chat: agent-triggered handoff to a new chat

rsolmano:
## What

You can now tell the agent **"ok, implement in new session"** and it starts a new chat — focused, seeded with a prepared kickoff prompt, and already working.

The agent calls the new host-owned **`start_new_chat`** tool (`{ title?, prompt }`). The host creates a sibling session in the same workspace (inheriting the caller's model + thinking effort, named at birth via pi's `setSessionName`), **broadcasts `session.created`**, then fires the kickoff prompt. The client whose active workspace matches opens + focuses the new tab; every other client gets a reopenable, runtime-alive history row (no tab spam). A rejected kickoff (bad model, missing key) fails the tool call loud, so the calling agent reports it in-conversation — never a silent empty chat.

The tool description carries the context contract: the new session has no memory of the calling one, so the prompt must be self-contained — preferably pointing at a task-spec / handoff doc under `.thinkrail/context/`.

## How

Follows three established patterns, one per ring:

- **contracts** — protocol **v39**: `session.created` channel + `SessionCreatedPayload` (the `session.deleted` sibling — a non-replayable domain *event*, not a snapshot), plus `StartNewChatArgs`/`StartNewChatDetails`.
- **server/agent** — `startNewChat.ts`: the tool + the host-installed `setStartNewChatHandler` seam (the `resolve_comment` delegation pattern — the agent module gains no new deps). The tool reads its own session id / model / effort from pi's `ExtensionContext`. `CreateSessionInput.title` → `setSessionName`; new `getLiveSessionSummary` read for the broadcast.
- **server/host** — `startNewChatFromOrigin` mirrors the `session.create` handler (workspace resolve → scratch-dir seed → create → `chat_started` track), then **broadcasts before awaiting the kickoff's accept-ack** (`ackSend(promptSession(…))`): per-socket FIFO guarantees the push precedes the new session's `pi.event` stream, so every client builds the runtime before the prompt's `message_start` arrives. Unlike review sends this is deliberately *not* detached — a tool call taking seconds is normal, and the error path stays in front of the user.
- **web** — transport subscription (+ `NON_REPLAYABLE_CHANNELS` classification: replaying a past creation could resurrect a since-deleted chat), the idempotent + tombstone-checked `applySessionCreated` store fold, and a passive `StartNewChatCard` receipt (primary prominence; the tab opens via the fold, so no interactive Open action).

Module SPECs updated across contracts / agent / host / transport / store / chat-tools (spec leads code).

## Tests

- **`start_new_chat` tool suite** — seam delegation carries `{provider,id}` only (never the full `Model` with `baseUrl`), empty-prompt fail-fast, seam-rejection propagation, fail-loud default handler.
- **Host compose integration test** (faux provider, real workspace registry): the `session.created` publish carries the titled live summary and lands **before** every pi event of the new session; unknown origin fails loud and publishes nothing.
- **Store fold matrix** — active workspace → tab + focus (+ nav tick); non-active → runtime-alive history row, no tab; idempotent; tombstone-checked.
- **Transport replay classification** — event channels are never replayed to late subscribers.
- **`@agent` e2e** (`new-chat-handoff.live.spec.ts`, on-demand suite) — **passed against a real provider**: the model calls the tool, a second chat tab opens focused and titled, the host-fired kickoff prompt renders as a user message and runs to completion, the origin chat shows the receipt card.

Gates: typecheck / lint / check:deps / check:seams / unit suites green. No-agent e2e run: every failure reproduced on the pristine tree (machine-environmental: layout-drag ×3, terminal-survive-navigation ×1, plus shard-contention flakes) — none in areas this change touches.

## Deferred (recorded in the task-spec)

Cross-worktree handoff (new branch/worktree), a brainstorming-skill mention of the tool, unifying review-send chat creation onto the same broadcast, and phrasing-discovery tuning (the e2e names the tool explicitly to stay deterministic).

### DIFF
--- a/apps/web/src/chat/tools/SPEC.md
+++ b/apps/web/src/chat/tools/SPEC.md
@@ -53,6 +53,12 @@ registration runs once when the chat module mounts. Unregistered tools fall back
     color). Shown up front for every recommended option, not gated on selection: more discoverable than
     a tooltip and, being ordinary visible text, it reads on touch and for AT without a popover.
     **Active card only** — the resolved record shows selections only, no rationale.
+- **`StartNewChatCard`** — the in-transcript receipt for the host-owned `start_new_chat` handoff tool
+  (capability: the server's `agent/startNewChat` + the host compose — see [[submodule-server-agent]]):
+  the new chat's title + the kickoff prompt it was started with. **Primary** — a handoff is a
+  user-visible outcome, not plumbing, so it escapes the activity fold (the collapsed header's summary is
+  the new chat's title). The card is deliberately passive: the tab itself opens + focuses via the
+  `session.created` store fold, so the card carries no interactive Open action.
 - **`visualize/`** — `VisualizationCard` dispatches on `args.type` to `DiagramCard` (mermaid → themed
   SVG via the **lazy-loaded** `mermaid`, source fallback on parse error) and `ComparisonCard` (option
   cards with pros/cons + `recommended` highlight); shared `MermaidView` re-renders on `[data-theme]`
--- a/apps/web/src/chat/tools/StartNewChatCard.tsx
+++ b/apps/web/src/chat/tools/StartNewChatCard.tsx
@@ -1,0 +1,33 @@
+import { MessageSquarePlus } from "lucide-react";
+import type { ReactNode } from "react";
+import type { ToolRenderProps } from "../toolRegistry";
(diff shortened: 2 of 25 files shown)


rsolmano:
Closing — changed direction for now.

PR #259, opened 2026-08-19
The /blog site + the first blog post ("Introducing ThinkRail")

makingthematrix:
## Summary

This PR introduces a subsite thinkrail.ai/blog where we can publish blog posts.

[**A short video recording of the blog**](https://drive.google.com/file/d/1Z-LflKewzhxdC61XO2MotCqvlypWN1y7/view?usp=sharing)

## Checklist


rsolmano (commented):
## Review: blog subsite (styling/font reuse + complexity)

Focus per review request: **(1)** does the blog reuse the website's styling/fonts/shared assets, **(2)** is there unnecessary complication. Findings were verified against a real build of this branch (`vite build` + `build-blog.ts`, served via `vite preview`, checked in a browser at desktop + 375px mobile viewports, plus adversarial rebuilds: `TZ=America/Los_Angeles`, malformed frontmatter).

**Headline: the blog doesn't actually reuse the site's fonts or palette.** With a light-OS browser: landing page → light theme, 23 self-hosted font faces loaded; blog → dark theme (inverted detection), **0 font faces loaded** (the copied CSS ships unresolved `@fontsource` package imports). The palette is a second, already-drifted copy rather than shared tokens, and the templates duplicate the theme bootstrap the landing page already has — with a `localStorage` key that splits the user's theme choice between the two halves of the site.

On complexity: the build script carries dead output (`posts.json`, `coverImage`), ~80 lines of fail-open fallbacks that turn content errors into silent green deploys, and 29 preloaded Shiki languages for a one-post blog.

| # | Priority | Finding | Location |
|---|----------|---------|----------|
| 1 | P1 | Self-hosted fonts never load (unresolved `@import`s in shipped CSS) | `blog.css:71` |
| 2 | P1 | Palette duplicated, already drifted from `styles.css` | `blog.css:89` |
| 3 | P1 | System theme detection inverted (both templates) | `index-template.html:15`, `post-template.html:14` |
| 4 | P1 | Fail-open build: broken post ⇒ green deploy without it | `build-blog.ts:291` |
| 5 | P1 | Blanket Biome exclusion hides 12 errors incl. a11y | `biome.json:15` |
| 6 | P1 | cmd install instruction wrong; fences render as inline code | post `index.md:21` |
| 7 | P1 | `github-dark` Shiki colors on white in light mode | `build-blog.ts:53` |
| 8 | P1 | No analytics on `/blog` (decision: it must be counted) | `post-template.html:8` |
| 9 | P2 | Footer wrong year/owner/license vs site | `index-template.html:45` |
| 10 | P2 | Post date off-by-one depending on build TZ | `build-blog.ts:172` |
| 11 | P2 | Video iframes lose 16:9 on mobile | `blog.css:535` |
| 12 | P2 | `posts.json`/`coverImage` are dead output | `build-blog.ts:315` |
| 13 | P2 | SPEC references deleted `content/blog/BLOG.md` | `SPEC.md:172` |

Each inline comment carries a concrete suggested fix. Several findings share one root cause — the blog being a parallel mini-site instead of reusing the site's build — so a shared base stylesheet + a small Vite-built blog entry (fonts, tokens, theme bootstrap, analytics) resolves #1, #2, #3, #8 and most of #9 together.

apps/website/src/blog/blog.css
```diff
@@ -0,0 +1,630 @@
...
+ * ## File Structure
+ *
+ * - Fonts: Self-hosted via @fontsource-variable packages
+ * - Theme tokens: Defined in :root and [data-theme] blocks
+ * - Layout: Mobile-first responsive design
+ * - Components: Header, index cards, post content, navigation, footer
+ */
+
+/* Self-hosted fonts — same as main site */
+@import "@fontsource-variable/geist";
```
rsolmano: **[P1] The blog's self-hosted fonts never load — these `@import`s are unresolved in the deployed CSS**

`build-blog.ts` copies this stylesheet verbatim to `dist/blog/blog.css` *after* `vite build` finishes, so Vite never resolves these bare package specifiers. The shipped CSS still contains `@import "@fontsource-variable/geist"`, which the browser resolves as a relative URL (`/blog/@fontsource-variable/geist`) — a 404 on GitHub Pages.

Verified at runtime: the landing page loads 23 self-hosted font faces; the blog loads **0** (`document.fonts` is empty) and renders in system-font fallbacks. So the blog does not actually reuse the site's Geist/JetBrains Mono/Orbitron faces, which is the point of the shared catalog pins (SPEC: "Fonts are self-hosted; the site makes no external font request").

**Suggested fix:** process `blog.css` through Vite instead of copying it — e.g. a small `vite build` call inside `build-blog.ts` with `blog.css` as input and a fixed output name (or emit a shared base stylesheet from the main build and `<link>` it from the templates). Then add a build assertion that `dist/blog/blog.css` contains no bare `@import "@fontsource` specifiers, so this can't regress silently.

apps/website/src/blog/blog.css
```diff
@@ -0,0 +1,630 @@
...
+	/* Font stacks (same as main site) */
+	--font-sans: "Geist Variable", "Geist", -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif;
+	--font-mono:
+		"JetBrains Mono Variable", "JetBrains Mono", "Fira Code", "SF Mono", "Cascadia Code", monospace;
+	--font-display:
+		"Orbitron Variable", "Orbitron", "Geist Variable", "Geist", -apple-system, BlinkMacSystemFont,
+		"Segoe UI", sans-serif;
+
+	/* Blog-specific semantic tokens — override these in [data-theme] blocks */
+	--blog-bg: var(--container-workspace-bg, #09090b);
```
rsolmano: **[P1] These tokens are a duplicated palette, not reuse — and they've already drifted**

The generated blog pages never load `styles.css`, so the `var(--container-workspace-bg, …)` fallback chain doesn't "inherit from the main site's palette" (as the header comment claims) — the values come from the `[data-theme]` blocks re-declared below in this file, a second copy of the palette.

The copy has already drifted: the site's light theme is workspace `#ffffff` / header `#f2f3f5` (`styles.css` `[data-theme="light"]`), while this file reverses them (workspace `#f2f3f5` / header `#ffffff`).

**Suggested fix:** extract the font stacks + theme palettes into one shared base stylesheet consumed by both the landing page and the blog (which also solves the font-loading issue above). `blog.css` should keep only the blog-specific aliases (`--blog-*`) and layout rules. The site already pins palette copies with tests (`src/palette.test.ts`, `src/fonts.test.ts`) — a second untested copy inside the same app works against that setup.

apps/website/src/blog/index-template.html
```diff
@@ -0,0 +1,63 @@
...
+	<title>Blog — ThinkRail</title>
+	<meta name="description" content="Updates, tutorials, and insights from the ThinkRail team.">
+	<link rel="icon" type="image/svg+xml" href="../favicon.svg">
+	<link rel="stylesheet" href="./blog.css">
+	<script>
+		// Blocking theme init — runs before first paint to prevent FOUC.
+		// Priority: localStorage > system preference > dark default.
+		(function() {
+			var stored = localStorage.getItem('blog-theme');
+			var theme = stored || (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'light' : 'dark');
```
rsolmano: **[P1] System theme detection is inverted: OS-dark gets the light theme and vice versa**

When `prefers-color-scheme: dark` matches this picks `'light'`. Verified at runtime with a light OS preference: the landing page opens **light**, then clicking through to the blog flips to **dark**.

```suggestion
			var theme = stored || (window.matchMedia('(prefers-color-scheme: light)').matches ? 'light' : 'dark');
```

Two adjacent reuse points while you're here:
- The landing page already solved this exact bootstrap (`index.html` head script + `main.ts`) with the key `thinkrail-site-theme`. Using a separate `blog-theme` key means a visitor's choice on the main site is ignored on the blog and vice versa — reusing the same key makes the preference follow them across Home ↔ Blog.
- `main.ts` also live-follows OS theme changes when no explicit choice is saved; the blog could get that for free from a shared snippet.

apps/website/src/blog/post-template.html
```diff
@@ -0,0 +1,70 @@
...
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>{{title}} — ThinkRail Blog</title>
+	<link rel="icon" type="image/svg+xml" href="../favicon.svg">
+	<link rel="stylesheet" href="./blog.css">
+	<script>
+		// Blocking theme init — runs before first paint to prevent FOUC.
+		// Priority: localStorage > system preference > dark default.
+		(function() {
+			var stored = localStorage.getItem('blog-theme');
+			var theme = stored || (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'light' : 'dark');
```
rsolmano: **[P1] Same inverted system-theme check as `index-template.html` — this template's copy**

```suggestion
			var theme = stored || (window.matchMedia('(prefers-color-scheme: light)').matches ? 'light' : 'dark');
```

(See the comment on `index-template.html` line 15 — same fix, and consider extracting this duplicated head-script/header/footer into one shared partial that `build-blog.ts` injects into both templates, so the two copies can't drift.)

apps/website/scripts/build-blog.ts
```diff
@@ -0,0 +1,430 @@
...
+		readingTime: number;
+	}> = [];
+
+	for (const dir of postDirs) {
+		try {
+			const { frontmatter, html, readingTime } = await parsePost(dir, hl);
+			const resolvedHtml = resolveImagePaths(html, frontmatter.slug);
+			posts.push({ dir, frontmatter, html: resolvedHtml, readingTime });
+		} catch (err) {
+			console.error(`   ⚠️  Skipping ${basename(dir)}: ${err}`);
```
rsolmano: **[P1] Fail-open build: a broken post silently vanishes from the deployed site**

Verified: removing `title:` from the current post's frontmatter still exits `0` and deploys an **empty blog index** — CI stays green while the post disappears. The same fail-open pattern repeats below for a missing post template (~60 lines of embedded fallback HTML), a missing index template, and a missing `blog.css` (deploying an unstyled blog).

For a static marketing site where deploy happens on push to `main`, a content/setup error should fail the build loudly, not degrade silently.

**Suggested fix:** drop this `try/catch` and the three fallback branches — let `parsePost`/`readFile` errors propagate to the existing top-level `build().catch(...)` which already exits non-zero. Only genuinely optional resources (e.g. a post without an `images/` dir, already handled by `existsSync`) deserve a recovery path. This deletes ~80 lines and makes the pipeline strictly simpler.

biome.json
```diff
@@ -7,7 +7,13 @@
 	},
 	"files": {
 		"ignoreUnknown": true,
-		"includes": ["**", "!designs", "!**/.claude", "!apps/web/src/styles/generated"]
+		"includes": [
+			"**",
+			"!designs",
+			"!**/.claude",
+			"!apps/web/src/styles/generated",
+			"!apps/website/src/blog/*-template.html"
```
rsolmano: **[P1] This exclusion silently removes lint/format/a11y coverage from the new templates**

Removing it locally surfaces 12 errors + 6 warnings in the two templates — including `lint/a11y/noSvgWithoutTitle` on the theme-toggle icons and the `{{placeholder}}` parse errors this exclusion was presumably added for.

Biome's own diagnostic names the fix for the placeholders: enable `html.parser.interpolation` (an `overrides` entry scoped to `apps/website/src/blog/*.html` keeps it local), then make the templates conform. The repo's convention treats gate suppressions as design signals (AGENTS.md "No silent suppressions"), and `styles.css`'s inline `<script>` handling followed the same rule — the templates shouldn't be the exception.

**Suggested fix:** replace this exclusion with:

```jsonc
"html": { "parser": { "interpolation": true } }
```

(globally or via an override), fix the reported a11y/style issues, and keep the templates linted.

apps/website/content/blog/2026-08-19-introducing-thinkrail/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+ThinkRail is a web-based graphical interface for the [Pi Coding Agent](https://pi.dev/), designed to complement its powerful command-line workflow. Pi’s strength lies in its flexibility: it is a minimal harness that lets you customize extensions and skills, and configure your working environment to fit your exact needs. This power, however, comes at a cost. New users often find themselves spending significant time learning how to navigate Pi, and even experienced users may find the workflow unnecessarily clunky.
+
+We believe ThinkRail can make Pi more accessible without compromising its power. It provides a visual layer on top of Pi, making it easier to onboard and become productive while preserving all of Pi’s depth and extensibility. What's more, ThinkRail’s integration with Git and GitHub also makes it easier to collaborate on projects in real time.
+
+## Great, how can I get it?
+
+ThinkRail ships as a single self-contained executable. To install it, open the terminal and use one of the following commands:
+
+1. **macOS/Linux** and **Windows** from cmd:
```
rsolmano: **[P1] Command Prompt instruction is wrong — `install.sh | bash` doesn't run in cmd**

The site's install contract (hero picker + Install section in `index.html`) is: native Windows shells (PowerShell **and** Command Prompt) use `install.ps1` via `powershell -c "irm … | iex"`; only **WSL** uses `install.sh`. As written, a cmd user pastes a bash pipeline into a shell that can't run it.

**Suggested fix:**

````markdown
1. **macOS/Linux** (and Windows via **WSL**):

   ```
   curl -fsSL https://raw.githubusercontent.com/JetBrains/thinkrail/main/install.sh | bash
   ```

2. **Windows** (PowerShell or Command Prompt):

   ```
   powershell -c "irm https://raw.githubusercontent.com/JetBrains/thinkrail/main/install.ps1 | iex"
   ```
````

Also note: triple backticks on one line are parsed as *inline* code, not fenced blocks (verified in the generated HTML — both commands render as `<code>` spans). Fences need to sit on their own lines, as above, to get the styled code blocks.

File: apps/website/scripts/build-blog.ts [hunk 8]
```diff
@@ -0,0 +1,430 @@
...
+	"kotlin",
+	"sql",
+	"graphql",
+	"docker",
+	"diff",
+	"plaintext",
+] as const;
+
+// Theme for syntax highlighting (matches the blog's dark theme)
+const SHIKI_THEME = "github-dark";
```
rsolmano: **[P1] Hardcoded `github-dark` makes future code blocks unreadable in the blog's light theme**

Shiki emits inline dark-theme foreground colors (e.g. `color:#79B8FF`), and the renderer below strips only `background-color` so `blog.css` controls the background — which in light mode is `--elevated: #ffffff`. Result: light-gray-on-white code at roughly 2:1 contrast.

This is latent today (the current post's one-line backticks render as inline code, not Shiki blocks), but it breaks the first post that ships a real fenced code block, and there's no gate that would catch it.

**Suggested fix:** use Shiki's dual-theme output (`themes: { light: "github-light", dark: "github-dark" }`), which emits both palettes via CSS variables, and switch them under `[data-theme]` in `blog.css` — or keep code blocks on a dark background in both page themes, like the landing page's terminal. Related simplification: 29 preloaded languages for a marketing blog is a lot; `plaintext/bash/powershell/typescript/json` covers the foreseeable posts, and shiki can lazy-load the rest.

apps/website/src/blog/post-template.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>{{title}} — ThinkRail Blog</title>
+	<link rel="icon" type="image/svg+xml" href="../favicon.svg">
+	<link rel="stylesheet" href="./blog.css">
```
rsolmano: **[P1] Blog pages ship no analytics — `/blog` pageviews are invisible**

Decision from review: **`/blog` must be counted in the site's analytics.** These generated pages load only `blog.css` and two inline theme scripts; they never run the site entry that initializes PostHog + GTM (`main.ts` → `initAnalytics()` / `initGtm()`), so every blog visit — including traffic that lands directly on a post — goes unrecorded.

**Suggested fix:** add a small Vite-built blog entry (e.g. `src/blog/blog.ts` importing `initAnalytics` + `initGtm`, plus the shared theme bootstrap from the comment on line 14) emitted with a fixed filename, and `<script type="module" src="./blog.js">` it from both templates. The existing hostname gate (`analyticsConfig` / `gtmConfig`) already keeps dev/preview silent, so nothing else is needed — and the SPEC's Analytics section stays true for the whole site.

(Same applies to `index-template.html`.)

apps/website/src/blog/index-template.html
```diff
@@ -0,0 +1,63 @@
...
+			<p class="blog-index-subtitle">Updates, tutorials, and insights from the ThinkRail team.</p>
+		</header>
+
+		<div class="blog-posts-list">
+			{{posts}}
+		</div>
+	</main>
+
+	<footer class="blog-footer">
+		<p>© 2025 ThinkRail. All rights reserved.</p>
```
rsolmano: **[P2] Footer contradicts the site's: wrong year, wrong owner, wrong license story**

The landing page's status bar says `Copyright © 2026 JetBrains InnovationHub · Apache License 2.0`; this says `© 2025 ThinkRail. All rights reserved.` — stale year and "All rights reserved" on an Apache-2.0 project.

**Suggested fix:** reuse the site's footer line verbatim:

```suggestion
		<p>Copyright © 2026 <a href="https://www.jetbrains.com/" target="_blank" rel="noopener noreferrer">JetBrains</a> InnovationHub · Apache License 2.0</p>
```

(Same in `post-template.html` line 52 — one more reason to extract the shared header/footer partial.)

apps/website/scripts/build-blog.ts
```diff
@@ -0,0 +1,430 @@
...
+
+	// Validate required frontmatter
+	if (!data.title || !data.slug || !data.date) {
+		throw new Error(`Post at ${postDir} missing required frontmatter (title, slug, date)`);
+	}
+
+	const frontmatter: PostFrontmatter = {
+		title: data.title,
+		slug: data.slug,
+		date: data.date,
```
rsolmano, line 287: **[P2] Post dates render off-by-one depending on the build machine's timezone**

`gray-matter` (js-yaml) parses the unquoted `date: 2026-08-19` as a `Date` at **UTC midnight** — so this field is a `Date`, not the `string` the `PostFrontmatter` type claims. `toLocaleDateString("en-US", …)` then formats it in the builder's local zone.

Verified: building with `TZ=America/Los_Angeles` renders this post as **"August 18, 2026"**; CI (UTC) says August 19. The `posts.json` date also serializes as `"2026-08-19T00:00:00.000Z"` instead of a plain calendar date.

**Suggested fix:** normalize at parse time — validate/format `data.date` to a `YYYY-MM-DD` string here, and format for display with `timeZone: "UTC"` (both here and in the index-card rendering below), so output is identical on every machine.

apps/website/src/blog/blog.css
```diff
@@ -0,0 +1,630 @@
...
+.blog-post-content img {
+	max-width: 100%;
+	height: auto;
+	margin: var(--space-lg) 0;
+	border-radius: 0.5rem;
+	border: 1px solid var(--blog-border);
+}
+
+.blog-post-content iframe {
+	max-width: 100%;
```
rsolmano: **[P2] Embedded videos lose their aspect ratio on mobile**

`max-width: 100%` shrinks the iframe's width but its HTML `height="360"` stays fixed — verified at a 375px viewport: both YouTube embeds render at ~343×360 (portrait-ish) instead of 16:9.

```suggestion
	width: 100%;
	height: auto;
	aspect-ratio: 16 / 9;
```

apps/website/scripts/build-blog.ts
```diff
@@ -0,0 +1,430 @@
...
+		coverImage: p.frontmatter.coverImage
+			? `./images/${p.frontmatter.slug}/${basename(p.frontmatter.coverImage)}`
+			: null,
+		tags: p.frontmatter.tags || [],
+		url: `./${p.frontmatter.slug}.html`,
+		readingTime: p.readingTime,
+	}));
+
+	// Write manifest
+	await writeFile(join(OUTPUT_DIR, "posts.json"), JSON.stringify(manifest, null, "\t"));
```
rsolmano: **[P2] `posts.json` (and `coverImage`) have no consumer — dead output**

The blog index is rendered directly from the in-memory `manifest` below; nothing in the repo fetches `dist/blog/posts.json`, and `coverImage` is threaded through the manifest but never rendered by either template.

**Suggested fix:** unless this is meant as a documented public feed (in which case the SPEC should say so — and an RSS/Atom feed would serve that goal better), drop the manifest file write and the `coverImage` plumbing. Smaller pipeline, no unused contract to maintain.

apps/website/SPEC.md
```diff
@@ -135,6 +135,48 @@ week, and `cancel-in-progress: false` so an in-flight publish finishes. Don't ad
...
+- Dark theme (default)
+- Light theme (`[data-theme="light"]`)
+- System preference detection (`prefers-color-scheme`)
+- Manual toggle via sun/moon button (persists in `localStorage`)
+
+See `src/blog/blog.css` header for the complete token reference and customization examples.
+
+### Content Author Guide
+
+`content/blog/BLOG.md` documents the frontmatter schema, Markdown features, image/video embedding,
```
rsolmano, line 171: **[P2] Dangling reference: `content/blog/BLOG.md` doesn't exist in this PR**

The "Removing other unnecessary files" commit deleted the author guide but this section still points to it — an `active` module spec referencing a missing file stops being ground truth.

**Suggested fix:** either restore `content/blog/BLOG.md` (the frontmatter schema/authoring guide is genuinely useful for non-dev post authors) or delete this subsection. If it's restored, the frontmatter schema should live only there, with this section linking to it, per the "say each thing once" spec rule.

rsolmano (changes requested):

makingthematrix:
I fixed the listed issues. @rsolmano : Please take a look especially at the added PostHog analytics - is this how it supposed to work?

PR #260, opened 2026-08-19
Refine ThinkRail themes and color palette

juliashilovaa:
Refines the ThinkRail theme system and aligns all themes with the updated product color palette.

Updates the ThinkRail palette with the new neutral shades and green accent colors.
Refines color values and contrast across the themes.
Consolidates the theme set to four supported themes only: Light, Dark, High Contrast Light, and High Contrast Dark.
Keeps all theme changes within the existing semantic color-token architecture.

## Also included

- Minor header and panel UI refinements.
- Terminal accessibility, contrast fixes.
- Tabs appearance redesign

Frame 25
Frame 26
Frame 28
Frame 30
Frame 31
Frame 32


apps/web/src/themes/bundled/darcula.theme.json
```diff
@@ -1,77 +0,0 @@
-{
```
rsolmano, line 1: why do you think it should be deleted? It's the only theme I use on a daily basis..

rsolmano (dismissed):

Issue #17, opened 2026-07-06
Show money estimation for session


rsolmano:
We show some money estimation already, let's check if it shows correct numbers

PR #261, opened 2026-08-19
design: delegation core + subagents (task specs for review)

SBOne-Kenobi:
## What this is

**Design review only — no implementation.** Two task-specs proposing the delegation framework ("session fabric") and subagents as its first consumer. Approving this design green-lights implementation; the specs are the artifact under review.

Background: subagent support was previously attempted by adopting a third-party pi extension and rolled back. Decision (recorded in the specs): build our own, and make **the framework the deliverable** — subagents are the extensibility check.

## Reading order

1. **`designs/TASK-delegation-core.md`** — the contract (`createChild` → `runQueued`/`runNow`, orthogonal axes, lineage, registry + lifecycle events), its semantics, and a decision log of how each choice was reached.
2. **`designs/TASK-subagent-support.md`** — the consumer layer: `Agent` + `get_subagent_result` tools, agent definitions (discovery/precedence/trust), definition → spawn mapping, web card + child transcript view.

Diagrams (mermaid) render inline on GitHub.

## What to scrutinize

- **The contract types** (`CreateChildSpec` / `ChildHandle` / `RunOptions` / `DelegationRunDetails`) — the core section of the first spec. Survived four adversarial design rounds; the decision log (Appendix B) records rejected alternatives (driver enum, pacing option, mega-spawnSession, per-pattern constructors).
- **The readiness rule**: V1 implements exactly one axis combination (hidden, non-interactive, fresh); everything else is typed and **loud-rejected** until its consumer lands (subsessions, branching, workflows — futures with deliberately unpinned UX).
- **Boundaries**: the pi-mirror firewall in `SessionOptions` (session-shaping mirrored, infrastructure never exposed); tree-not-DAG lineage (engine DAGs stay engine-side); the `WorkspaceProvider` isolation seam.
- **Trust posture** (second spec, decision 6): worktree agent definitions can never shadow built-in/personal names.

## Notes

- Task-specs normally live in `.thinkrail/context/` (gitignored scratch, per the spec-graph convention); they are moved into `designs/` for this review and will move back / be retired as implementation lands and decisions are promoted into module SPEC.md files.
- Specs are `status: draft` — the flip to `active` follows this review.

## Revision — review round 1 (cf4829c)

The boundary question raised in the review comment is resolved — **adopted in strengthened form**, after settling a new requirement with the owner: the layer must **work without ThinkRail, under vanilla pi**.

- **The delegation core itself becomes a portable package** (`packages/pi-delegation`, pi SDK as `peerDependency`, in-process children), with `packages/pi-subagents` as its first consumer — not a host-internal factory, and not a host-injected `SpawnBackend` either (that would demand a second pure-pi run loop; no reference implementation injects a spawner). Field research: core spec decision log **#17–21**.
- Host intimacies became explicit **embedder bindings** (`DelegationBindings`: delegation root, scope, parent resolution) — see the new "Portability & embedding" section.
- `DelegationRunDetails` stays **authored for the wire and mirrored into `contracts`** (mirror-never-import, the `pi-todos` DTO posture) — not re-exported from the package.
- Pure-pi V1 bar (owner-settled): the extension **loads + works under default tool rendering**; no pi-tui widget in V1.

### DIFF
--- a/designs/TASK-delegation-core.md
+++ b/designs/TASK-delegation-core.md
@@ -3,7 +3,7 @@ id: task-delegation-core
 type: task-spec
 status: draft
 title: "Delegation core — session fabric: spawn, fork, lineage, extension points"
-parent: submodule-server-agent
+parent: architecture
 references: [task-subagent-support]
 ---
 
@@ -11,8 +11,12 @@ references: [task-subagent-support]
 
 ## Summary
 
-A `delegation` sub-module in `packages/server/src/agent/`: the framework for creating agent
-sessions *from* agent sessions. One creation primitive with orthogonal axes; a handle that owns
+`packages/pi-delegation` — a **portable, pure-pi workspace package**: the framework for creating
+agent sessions *from* agent sessions. Children run **in-process** via the pi SDK (a
+`peerDependency`; the loading runtime supplies it), and the layer must **work without ThinkRail,
+under vanilla pi** — ThinkRail's server is one *embedder*, binding a small set of host facts
+("Portability & embedding" below; superseded: the original placement as a `packages/server`
+sub-module — decision log #17). One creation primitive with orthogonal axes; a handle that owns
 the run loop; lineage, a run registry, and lifecycle events as extension points. **V1 implements
 exactly one axis combination — subagents ([[task-subagent-support]], the extensibility check);**
 every other combination is typed in the contract and rejected loudly until its consumer lands.
@@ -36,7 +40,9 @@ will be built (by us and, eventually, by the user for their own workflows):
 
 ## The contract
 
-Host-side types on the `delegation` barrel; `DelegationRunDetails` lives in `packages/contracts`.
+Types on the `pi-delegation` barrel. `DelegationRunDetails` is **mirrored** into
+`packages/contracts` (mirrored, never imported — the `pi-todos` DTO posture, decision #20), which
+keeps the package host-free and the `web → contracts`-only invariant intact.
 
 ```ts
 // ── The service (= the barrel) ───────────────────────────────────────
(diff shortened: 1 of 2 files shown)


rsolmano:
## Decision #1 — proposing to discuss: portable package + spawn seam instead of a host-internal factory

Design-review question on `designs/TASK-subagent-support.md`, settled decision #1 (*"Host-owned bundled extension — an extension factory in `packages/server/src/agent` … not a separate workspace package; extraction stays possible"*).

**Proposal to consider:** the subagent layer (the `Agent` / `get_subagent_result` tools + the definitions loader + `DelegationRunDetails`) lives in its own workspace package (`packages/pi-subagents`, same posture as `pi-spec-graph` / `pi-todos`), consuming a narrow **spawn seam**. The delegation core stays host-owned exactly as `TASK-delegation-core.md` specifies — it becomes the seam's binding.

### Why this is cheap to decide now

- The core's own acceptance criterion already requires **"zero private child-assembly code in the tool layer"** — the tools touch nothing but the service contract. So the seam effectively already exists in this design; reversing decision #1 changes *where the files live and which types they import*, not the machinery. (`SpawnBackend` ≈ `createChild` + `runQueued` — a subset of the existing `DelegationService`.)
- The repo's documented tool pattern is already two-sided ("capability = pi extension, presentation = renderer joined by tool name"), so the web card / transcript-view side is unaffected.

### Why it matters

- Module boundaries are a top-priority requirement in this repo, and a package boundary is **enforced by construction**; "extraction stays possible" is a posture that decays as host coupling accretes. The delta is smallest at design time and grows once the layer lands inside `packages/server/src/agent/`.
- The rolled-back third-party adoption argued against *adopting* subprocess-spawning / process-wide-registry / TUI-coupled extensions — not against *authoring* a layer that already makes the opposite choices (in-process children via the core, per-session discovery, `partialResult` data instead of TUI components). The history doesn't bind this boundary decision.

### Mechanical consequences to spell out if we go this way

1. The package exports `createSubagentExtension(backend: SpawnBackend)`; the host binds the delegation core to it.
2. `DelegationRunDetails` is defined in the package and type-re-exported through `packages/contracts` (same types-only posture as the `pi-agent-core` / `pi-ai` devDeps), preserving the web→contracts-only invariant.
3. The definitions loader stays self-contained — discovery / precedence / trust as specced carry nothing host-specific today.

Not proposing any change to the delegation core spec's contract — the ask is only whether the subagent layer's *boundary* should be a package + seam rather than a host-internal factory, and it seems worth settling before the spec flips to `active`.

PR #267, opened 2026-08-20
feat(web): browser Back/Forward steps through in-app navigation history

### LINKED ISSUE #152: respect browser controls (back --- step back in history)

rsolmano:
Closes #152. Builds on the route model + validated restore that #262 shipped.

## What

Browser **Back/Forward now step through in-app navigation**: Welcome ↔ Project Home ↔ workspace ↔ exact chat, including chat-tab switches and chat closes. Until now every navigation used `history.replaceState`, so Back left the app.

## How

History entries are **locations, not tab state** — every entry is a route the existing restore contract can validate and re-enter, so Back/Forward simply arrives as an incoming fragment (`hashchange`) and rides #262's validated restore path unchanged. The new machinery only decides *when a navigation creates an entry*:

- **User-intent location changes push**, passive ones **replace** (coalescing) — one click = one history entry, whose fragment self-heals to the final location (e.g. entering a workspace pushes once; the auto-opened chat replaces into that same entry).
- The coordinator detects intent per store edge: a scope move, the active workspace's `navTickByWorkspace` advance, an attention **group navigation-clock** advance, or **removal of the attention-selected center tab** (a close moves the location at its optimistic document commit, before its deferred navigation count lands on host acceptance).
- An intent edge **arms a push**; the next fragment write consumes it. The arm survives unchanged-location syncs because **deferred stamped opens** (new-chat button, history reopen, cross-workspace jumps) count navigation at the click and land the location change only after an async round trip, deliberately without re-counting.
- **Adoption:** incoming fragments (startup, Back/Forward, address-bar edits) never re-push; the coordinator's own route-applying writes are masked from the intent test.
- `NavigationDriver` grows `push(fragment)` (`history.pushState` in the browser driver); native adapters map it onto their own history model.

Per the settled design (recorded in `apps/web/src/navigation/SPEC.md` → *History contract*):

- **Granularity:** scope moves **and** chat-tab switches create entries (full route fidelity).
- **Closed chats:** a chat's history entry survives its tab being closed — Back re-opens it via the normal validated restore; closing the active chat pushes the neighbor location, so Back after close returns to the closed chat. Deleted chats fall back canonically (workspace retained, chat intent consumed).
- **Accepted coarseness** (route contract unchanged): file/diff/doc tabs still serialize at workspace granularity; a Back onto a same-workspace workspace-level entry may canonicalize to the selected chat (one dead Back press, self-healing). Exact-file routes and a same-workspace fast path are recorded follow-ups.

## Tests

- `apps/web/src/navigation/restore.test.ts`: fake driver logs push vs replace; new coverage for chat-switch push, passive auto-open coalescing, Back/Forward adoption (never re-pushed), armed-push survival across the underivable-location hold, the attention-before-commit open window, the deferred stamped open, and close-as-push. 29 tests.
- New `e2e/history-navigation.spec.ts` (no-agent): Back/Forward through chat switches and scope moves with `page.goBack()`/`goForward()`; Back returning to a deep-linked chat; a closed chat re-opened by Back; a deleted chat's entry falling back to the workspace.

## Gates

`check:deps` · `check:seams` · `lint` · `typecheck` · unit tests · **full no-agent e2e suite (8 shards)** — all green.


rsolmano:
The binary e2e lane caught a real race (not an artifact-only issue): **Back to a just-closed chat** could be cancelled by the close's delayed `layout.replace` acknowledgement. The ack landed while the route's authoritative `workspace.list` read was still in flight, saw no attention-clock movement, counted itself as fresh navigation, and its `navTick` bump cancelled the pending route — the Back appeared dead and the hash flipped backward. Fast local machines never hit the window, which is why the dev lane passed.

Fixed in 34f744c2, three coordinated parts (rationale in the navigation / store / layout SPECs):

1. **Route adoption bumps the active workspace's `navTick` immediately** (masked), before the validation read — an incoming Back/Forward is user navigation, visible to request-stamped staleness checks.
2. **The layout close's `navigationWasOvertaken` also compares the workspace navTick captured at close request** (new `readNavigationTick` prop on `Workbench`), so a stale acceptance defers to the newer Back instead of cancelling it.
3. **`closeChatToHistory` no longer clears `routeChatTarget`** — the acceptance is an older click's echo; target lifecycle belongs to navigation supersession + reconciliation (covers the interleaving where the ack lands after the target installs).

Pinned forever by a deterministic e2e test that holds the `layout.replace` ack and the Back's `session.list` response and releases them in CI's order (reproduced the failure before the fix, passes after), plus a store unit test.

danyaberezun (dismissed):
One blocking Back/Forward race remains. I reproduced it with a focused coordinator test. I also left one history-arm correctness comment and two small non-blocking cleanups.

danyaberezun (dismissed):
One more blocking Back/Forward case is confirmed below.

danyaberezun (approved):

PR #272, opened 2026-08-21
web: declutter the specs tree rows

CommanderTvis:
The trailing role column repeated down every row and truncated to `SUBMOD…` in the narrow rail; it now appears on row hover/focus only, untruncated, with the role still in each row's aria-label. Spec titles render their name/summary separator as `·` instead of an em dash.

## Summary
Before
image

After
image

## Checklist

### DIFF
--- a/e2e/specs-panel.spec.ts
+++ b/e2e/specs-panel.spec.ts
@@ -26,6 +26,7 @@ test("Specs tab renders the worktree's spec tree and opens a spec as an editor t
 	await expect(child).toHaveAttribute("data-spec-role", "MODULE");
 	await expect(child.getByTestId("spec-role")).toBeHidden();
 	await child.hover();
+	await expect(child.getByTestId("spec-role")).toBeVisible();
 	await expect(child.getByTestId("spec-role")).toHaveText("MODULE");
 
 	expect(await child.getAttribute("data-status")).toBeNull();
--- a/e2e/specs-panel.spec.ts
+++ b/e2e/specs-panel.spec.ts
@@ -20,6 +20,7 @@ test("Specs tab renders the worktree's spec tree and opens a spec as an editor t
 	await expect(root).toHaveAttribute("data-spec-role", "Main spec");
 	await expect(root.getByTestId("spec-role")).toBeHidden();
 	await root.hover();
+	await expect(root.getByTestId("spec-role")).toBeVisible();
 	await expect(root.getByTestId("spec-role")).toHaveText("Main spec");
 	await expect(child).toHaveAttribute("data-depth", "1");
 	await expect(child).toHaveAttribute("data-spec-role", "MODULE");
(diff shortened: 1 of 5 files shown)


e2e/specs-panel.spec.ts
```diff
@@ -18,10 +18,14 @@ test("Specs tab renders the worktree's spec tree and opens a spec as an editor t
 	await expect(root).toHaveAttribute("data-depth", "0");
 	await expect(root).toHaveAttribute("data-main-spec", "true");
 	await expect(root).toHaveAttribute("data-spec-role", "Main spec");
-	await expect(root).toContainText("Main spec");
+	await expect(root.getByTestId("spec-role")).toBeHidden();
+	await root.hover();
+	await expect(root.getByTestId("spec-role")).toHaveText("Main spec");
```
rsolmano, line 24: `toHaveText()` reads `textContent`, so it still passes while the role stays `display:none` — the hover reveal itself isn't asserted. Add a visibility check:

```suggestion
	await root.hover();
	await expect(root.getByTestId("spec-role")).toBeVisible();
	await expect(root.getByTestId("spec-role")).toHaveText("Main spec");
```

e2e/specs-panel.spec.ts
```diff
@@ -18,10 +18,14 @@ test("Specs tab renders the worktree's spec tree and opens a spec as an editor t
...
 	await expect(child).toHaveAttribute("data-depth", "1");
 	await expect(child).toHaveAttribute("data-spec-role", "MODULE");
-	await expect(child).toContainText("MODULE");
+	await expect(child.getByTestId("spec-role")).toBeHidden();
+	await child.hover();
+	await expect(child.getByTestId("spec-role")).toHaveText("MODULE");
```
rsolmano, line 30: Same here — assert the reveal, not just the text:

```suggestion
	await child.hover();
	await expect(child.getByTestId("spec-role")).toBeVisible();
	await expect(child.getByTestId("spec-role")).toHaveText("MODULE");
```

e2e/specs-panel.spec.ts
```diff
@@ -72,6 +76,8 @@ test("Specs tab renders the worktree's spec tree and opens a spec as an editor t
 	await expect(submodule).toBeVisible();
 	await expect(submodule).toHaveAttribute("data-depth", "2");
 	await expect(submodule).toHaveAttribute("data-spec-role", "SUBMODULE");
+	await submodule.hover();
+	await expect(submodule.getByTestId("spec-role")).toHaveText("SUBMODULE");
```
rsolmano, line 84: Same `toHaveText`-on-hidden gap, and this block also skips the pre-hover hidden check the earlier rows have:

```suggestion
	await expect(submodule.getByTestId("spec-role")).toBeHidden();
	await submodule.hover();
	await expect(submodule.getByTestId("spec-role")).toBeVisible();
	await expect(submodule.getByTestId("spec-role")).toHaveText("SUBMODULE");
```

rsolmano (changes requested):
Nice cleanup — verified locally that lint/typecheck/unit/e2e all pass on this branch. Requesting changes only for the two still-open Copilot test findings: `toHaveText()` passes on `display:none` elements, so the hover reveal isn't actually exercised. Ready-to-apply suggestions inline.

rsolmano (approved):
Thanks for the quick turnaround and the clean contribution, @CommanderTvis! 🙏 All suggestions applied, verified locally — LGTM.

juliashilovaa (approved):

PR #275, opened 2026-08-21
chat: mid-stream messages — pi-parity queueing, legible send modes, per-row queue editing, hard interrupt

rsolmano:
## Summary

Two stacked changes to how mid-stream messages behave in chat.

### 1. Queued/steered messages: vanilla-pi parity (fixes a live-reproduced defect)

The composer placeholder promises "Cmd/Ctrl+Enter to queue" — queueing worked (pi queues, doesn't interrupt, runs the follow-up after the turn), but a live repro showed the transcript rendering the queued message **twice**, with the optimistic copy **above the reply it didn't prompt**. Root cause: the store's user-echo dedup checked only the last turn, which misses whenever assistant content lands between the optimistic append and pi's canonical echo (followUp drain, steer window, tool calls, first-token latency). Live and reloaded transcripts disagreed.

Fix mirrors vanilla pi's interactive mode:

- A **streaming send (`steer`/`followUp`) no longer appends an optimistic transcript bubble.** The turn lands only via pi's canonical user `message_start` — at its true position, converging live with hydrated. The defect disappears structurally.
- Queued texts render in a **pending strip** above the composer (pi's pending-messages area), folded from pi's `queue_update` event (typed in contracts, previously unconsumed).
- **Abort restores the queue to the composer** (pi's Escape parity); a **rejected** queued send restores its text to the draft.
- **Reconnect snapshot:** live session summaries carry pi's queue (`SessionQueueState`), so a client attaching mid-run renders messages queued before it connected.
- Idle sends keep the optimistic echo + last-turn dedup — sufficient once nothing intervenes before the echo.

### 2. Streaming send-mode legibility + per-row queue management + hard interrupt (field feedback on #1)

- **Split send control while streaming:** a send-options menu names each mode with its meaning and shortcut — *Steer — delivers at the agent's next step (Enter)* / *Queue — runs after the agent finishes (Cmd/Ctrl+Enter)* / *Interrupt — stops the current response and sends now (Cmd/Ctrl+Shift+Enter)*. Rows are actions, never a sticky mode. The placeholder states meanings, not just key names (steer ≠ interrupt was the #1 confusion).
- **Pending strip rows get per-message edit + remove** via the new `session.removeQueued { kind, index }` wire method. pi's queue API is all-or-nothing, so the host emulates per-item removal (clearQueue → re-queue keepers) with a **no-loss idle fallback** (keepers that would park on a settled session deliver via the idle path instead).
- **Hard interrupt:** await `session.abort` (ack = idle), then an ordinary send; the partial reply stays marked aborted.

New wire surface: `session.clearQueue`, `session.removeQueued`, `SessionSummary.queue` (protocol v46 → v48).

Accepted trade-off (recorded in `chat/SPEC.md`): `queue_update` is text-only, so a queued image attachment shows no chip in the strip; its image blocks render when the message lands in the transcript.

Design records: brainstormed task-specs promoted into `apps/web/src/store/SPEC.md` (echo contract), `apps/web/src/chat/SPEC.md` (strip rows, split send, interrupt), `packages/server/src/agent/SPEC.md` + `packages/contracts/SPEC.md` (`clearQueue`/`removeQueued`, queue snapshot).

## Related issues

None — found by reproducing the placeholder's claim, then field-testing the first fix.

## Verification

- Unit: `queue_update` fold + hydration seeding (web store); mid-stream queue snapshot, `clearQueueSession`, `removeQueuedSession` (siblings kept, out-of-range no-op, idle no-loss fallback) against a faux provider (server).
- Permanent `@agent` spec **`e2e/queue.live.spec.ts`** against a real provider: queue mid-stream → strip row appears, first turn uninterrupted, queued turn runs at canonical position, exactly 2 bubbles; row remove + row edit (text back in composer); interrupt aborts and the new reply arrives.
- Full `@agent` suite (`THINKRAIL_E2E_MODEL=openai-codex/gpt-5.4-mini`): 42 passed, 6 failed — **the same 6 fail identically on `origin/main`** with that model (model-sensitivity of specs tuned for the pinned claude default; verified on a detached checkout). No regressions.

## Checklist

- [x] E2E suite passes for app-affecting changes (`bun run e2e` — 252 passed / 8 shards; agent behavior covered by the `@agent` runs above)

### DIFF
--- a/apps/web/src/chat/ChatView.tsx
+++ b/apps/web/src/chat/ChatView.tsx
@@ -1,6 +1,7 @@
 import type {
 	AskUserQuestionResult,
 	PromptHit,
+	QueueLane,
 	SlashCommandInfo,
 	TemplateInfo,
 	ThinkingLevel,
@@ -285,7 +286,11 @@ export default function ChatView({
 		composerRef.current?.refocus();
 	};
 
-	const onSubmit = (text: string, attachments: ChatAttachment[], behavior: SubmitBehavior) => {
+	const performSend = (
+		text: string,
+		attachments: ChatAttachment[],
+		behavior: Exclude<SubmitBehavior, "interrupt">,
+	) => {
 		const queued = behavior !== "send";
 		if (!queued && (text || attachments.length > 0))
 			useAppStore.getState().appendUserMessage(sessionId, text, attachments);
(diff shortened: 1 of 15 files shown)


rsolmano:
### Before / after screenshots (real provider, live sessions)

All captured by driving the real UI against a real agent (`gpt-5.4-mini`): first turn = "count from 1 to N", then `Cmd/Ctrl+Enter` with a follow-up while it streams.

#### Before — the defect (main, pre-PR)

| Mid-stream | Settled |
|---|---|
| before: queued bubble mispositioned above the streaming reply | before: the queued message renders twice |

*Mid-stream:* the optimistic bubble sits **above** the reply it didn't prompt (and the canonical copy already landed below — duplicate visible while streaming). *Settled:* the queued message appears **twice** — once mispositioned above the count reply, once at its true position.

#### After — this PR

| Pending strip while streaming | Send-options menu | Settled |
|---|---|---|
| after: pending strip row with edit/remove | after: send menu naming steer/queue/interrupt with meanings | after: canonical transcript, single bubble |

*Streaming:* no premature transcript bubble — the queued message shows as a **pending-strip row** (`Follow-up: …`) with per-message **edit/remove** actions, and the **send menu** documents Steer / Queue / Interrupt with their meanings and shortcuts. *Settled:* exactly one bubble, at canonical position (matches what a reload shows).

Images live on the throwaway `assets/pr-275-screens` ref — delete it after merge.

danyaberezun (approved):

PR #276, opened 2026-08-21
web: clearer tab-strip affordances and faster tooltips

CommanderTvis:
- Show custom tooltips for icons with unclear for a new user meaning with less delay because native ones can have uncontrollable and long delay
image

- Show tab scroll and search controls only while the tabs actually overflow
image

- Show the fold control only while the side holds more than one group
image

- Offer unplaced tools from a menu on the side strip itself, scoped to that side, not only from a tab's context menu
image

- Create a terminal from that menu, in the group it was opened from

### DIFF
--- a/apps/web/src/shell/layout/Workbench.tsx
+++ b/apps/web/src/shell/layout/Workbench.tsx
@@ -23,6 +23,7 @@ import {
 	RiLayoutLeftLine as PanelLeftOpen,
 	RiLayoutRightLine as PanelRightOpen,
 	RiLayout2Line as PanelsTopLeft,
+	RiAddLine as Plus,
 	RiBookOpenFill,
 	RiBookOpenLine,
 	RiChat2Fill,
@@ -37,6 +38,7 @@ import {
 	RiGitPullRequestFill,
 	RiLayout2Fill,
 	RiTerminalBoxFill,
+	RiSearchLine as Search,
 	RiTerminalBoxLine as SquareTerminal,
 	RiCloseLine as X,
 } from "@remixicon/react";
@@ -86,6 +88,7 @@ import {
 	ResizablePanel,
 	ResizablePanelGroup,
 } from "../../components/ui/resizable";
+import { IconTooltip } from "../../components/ui/tooltip";
 import {
 	DOUBLE_CLICK_SETTLE_MS,
 	type LayoutAttention,
@@ -113,7 +116,6 @@ import {
 	isLayoutUnavailable,
 	keepPreview,
 	LAYOUT_LIMITS,
-	LAYOUT_TOOLS,
 	type LayoutGroupLocation,
 	type LayoutMutationResult,
 	type LayoutOperationResult,
@@ -135,6 +137,8 @@ import {
 	showSide,
 	splitCenterGroup,
 	toolTab,
+	unplacedTools,
+	unplacedToolsForSide,
 } from "./model";
 
 export interface LayoutTabFocusRequest {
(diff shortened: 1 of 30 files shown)


CommanderTvis:
By the way, most of issues fixed are exactly ones that I've noticed when being very new to Thinkrail: unobvious icon controls should self-explain.

CommanderTvis:
One more BTW. Would you be open to a chat channel for ThinkRail (Discord/Telegram/whatever's least effort)?

Most of things in this PR is small stuff, not worth reporting them as issues. 

The other half is the opposite end. There's a large experiment like the ACP migration I've been poking at https://github.com/CommanderTvis/thinkrail/tree/acp-migration. Explaining where that's going is a conversation, not an issue or GH discussion.

danyaberezun (changes requested):
Thank you for the contribution. There are just two small things to fix before merging.

**The new viewport assertion is flaky.** `layout.spec.ts:426-429` fails about 1 in 3 for me, and it took down the full no-agent suite (everything else green). The two `setViewportSize` calls are back to back, so Chromium coalesces them — the ResizeObserver never sees the wide state, `setOverflowOpen(false)` never runs, and the popover is still open when the trigger remounts. Adding the intermediate wait passed 5/5:

```ts
await page.setViewportSize({ width: 1280, height: 720 });
await expect(searchTabs).toHaveCount(0);
await page.setViewportSize({ width: 620, height: 800 });
```

(Deriving the reset from `scrollable` in its own effect instead of from inside `measure()` would make it race-free either way.)

**`<span className="flex">` is in 11 places now** — always as `IconTooltip`'s direct child, always for one of the two reasons the SPEC names. Could it move inside `IconTooltip` behind a prop? That span is what keeps the tooltip's `data-state` off the popover trigger — the bug you already pinned in `review.spec.ts` — and right now the only thing stopping a 12th call site from reintroducing it is remembering to read the SPEC.

Rest reads well. The side-scoped `+` menu is a real improvement, and dropping the tools block from center tab menus is the right call.

danyaberezun (dismissed):
Requesting changes — please address the blocking inline finding.

apps/web/src/shell/layout/Workbench.tsx
```diff
@@ -977,12 +1030,15 @@ function WorkbenchTab({
 						})}
 					</>
 				) : null}
-				{missingTools.length > 0 ? (
+				{/* Tools live in the side regions, and the side strip's own menu offers the missing ones there.
```
danyaberezun: Blocking:
Problem: This new multi-line implementation rationale violates the repository’s explicit near-zero-comments invariant; the same decision is already recorded in the owning layout spec.
Failure scenario: The spec and this duplicated prose can drift independently when tool recovery changes, leaving reviewers and maintainers with conflicting sources of truth.
Suggested fix: Remove the comment and keep the rationale solely in `apps/web/src/shell/layout/SPEC.md`.

danyaberezun (changes requested):
Requesting changes — please address the blocking inline finding.

apps/web/src/shell/layout/Workbench.tsx
CommanderTvis: Clear, I was sending the PR before learning about this convention.

CommanderTvis:
One decision that is questionable: this PR makes currently custom tooltips only for icons that need them to be clear. It does not change all tooltips, which may be considered better because of consistency.

danyaberezun (dismissed):

rsolmano (dismissed):

CommanderTvis:
The failing tests look unrelated.

PR #279, opened 2026-08-23
fix(terminal): nudge PTY redraw on reattach so alt-screen TUIs reappear

CommanderTvis:
## Bug
Open `vim` (any full-screen TUI)
image

Visit another tab, go back to `vim` tab - and there's nothing
image

## Summary
- Reattaching to a terminal tab whose grid size didn't change is a no-op resize, so `TIOCSWINSZ` never raises `SIGWINCH` and a foreground full-screen app (vim, htop, an interactive CLI) is left showing the frozen pre-alt-screen buffer — the recorder deliberately never captures alt-screen content to replay in its place.
- `nudgePtyRedraw()` (`packages/server/src/terminal/ptyGrid.ts`) shrinks the column count by one and resizes back when the real resize on attach was a no-op, forcing two genuine `SIGWINCH` deliveries without moving the tracked grid. Wired into `terminalManager.ts`'s reattach branch.

## Test plan
- [x] `bun test packages/server/src/terminal/ptyGrid.test.ts packages/server/src/terminal/terminalManager.test.ts` — 28/28 pass
- [x] `turbo run typecheck` — clean
- [x] `biome check` — clean
- [x] Manually reproduced: opened `claude`/`vim` in a terminal tab, switched to another tab, switched back — TUI now redraws instead of showing a frozen shell prompt

### DIFF
--- a/packages/server/src/terminal/SPEC.md
+++ b/packages/server/src/terminal/SPEC.md
@@ -39,10 +39,30 @@ separate layout snapshot and merely references `tabKey`.
   its process is Theia's `Couldn't attach - can't find terminal with id`).
 - **`attachTerminal` is idempotent get-or-create** — the only way a PTY is born. No separate liveness call, so
   there is no window in which a client holds the only pointer to a running shell.
-- **PTY resizing is change-only.** Each live entry tracks the grid applied at spawn or by the last successful
-  resize. Attach and explicit resize call `IPty.resize` only when that grid changes, and failed calls do not
-  advance the tracked state. Even a same-grid resize can wake the shell through `SIGWINCH`; a redraw emitted
+- **Tracked-grid updates are change-only.** Each live entry tracks the grid applied at spawn or by the last
+  successful resize. Attach and explicit resize advance that grid only when it changes, and failed calls do not
+  advance it. A reattach may still perform a *transient* redraw nudge (below) that leaves the tracked grid
+  untouched. Even a same-grid resize can wake the shell through `SIGWINCH`; a redraw emitted
   after the attach snapshot can overwrite freshly replayed rows.
+- **Reattaching to an unchanged grid still nudges the foreground app** (`nudgePtyRedraw`, `ptyGrid.ts`):
+  `TIOCSWINSZ` to the same size is a kernel no-op — no `SIGWINCH`, so a full-screen alt-screen app (vim,
+  htop, an interactive CLI) left running behind a tab that was switched away and back gets no signal to
+  repaint, and the recorder never captured its alt-screen content to replay in its place (see below) — the
+  tab would otherwise show a frozen pre-alt-screen buffer. `attachTerminal`'s reattach branch resizes to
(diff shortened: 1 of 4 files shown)


packages/server/src/terminal/ptyGrid.ts
```diff
@@ -14,3 +14,43 @@ export function resizePtyIfChanged(pty: PtyResizer, current: PtyGrid, next: PtyG
...
+export function nudgePtyRedraw(
+	pty: PtyResizer,
+	current: PtyGrid,
+	options: NudgePtyRedrawOptions = {},
+): void {
+	const { cols, rows } = current;
+	const schedule = options.schedule ?? ((fn, ms) => setTimeout(fn, ms));
+	const isStillLive = options.isStillLive ?? (() => true);
+	const restore = (): void => {
+		if (isStillLive()) pty.resize(cols, rows);
```
rsolmano: Please guard this restore against a newer resize. I reproduced: attach at 46×21, resize to 38×21 during the delay, and the PTY ends at 46×21 while its tracked grid is 38×21. Retrying 38×21 is then a no-op. Checking that `current` still matches the captured grid fixes it and preserves the Vim redraw. Please add a regression test.

packages/server/src/terminal/ptyGrid.ts
```diff
@@ -14,3 +14,43 @@ export function resizePtyIfChanged(pty: PtyResizer, current: PtyGrid, next: PtyG
 	current.rows = next.rows;
 	return true;
 }
+
+// A same-size resize is a kernel no-op — TIOCSWINSZ only raises SIGWINCH on an actual change — so a
```
rsolmano: Please remove this block comment. `AGENTS.md` puts multi-line rationale in the owning spec, and this duplicates `terminal/SPEC.md`. If a pointer is needed, keep one short `see terminal/SPEC.md` note.

packages/server/src/terminal/SPEC.md
```diff
@@ -43,6 +43,23 @@ separate layout snapshot and merely references `tabKey`.
   resize. Attach and explicit resize call `IPty.resize` only when that grid changes, and failed calls do not
   advance the tracked state. Even a same-grid resize can wake the shell through `SIGWINCH`; a redraw emitted
   after the attach snapshot can overwrite freshly replayed rows.
+- **Reattaching to an unchanged grid still nudges the foreground app** (`nudgePtyRedraw`, `ptyGrid.ts`):
```
rsolmano, line 47: This contradicts the previous “resizing is change-only” rule. Please clarify that tracked grid updates are change-only, while reattach may perform a transient redraw nudge.

rsolmano (changes requested):
The original bug is fixed. Requesting changes for the verified resize race; two smaller comments cover code/spec consistency.

packages/server/src/terminal/ptyGrid.ts
CommanderTvis: Fixed. Convention noted, will self-review all subsequent contribs.

rsolmano (approved):

PR #282, opened 2026-08-24
fix(chat): Add note appears under each checked option in ask_user_question, multi-select included

OLavrik:
## Problem

Since #228 the "Add note" control in the ask_user_question card rendered below the
whole option list, visually attaching to the last option instead of the one it
belonged to — and multi-select questions had no note affordance at all, even though
the wire contract (`AskUserQuestionAnswer.notes`) always allowed notes on any answer.

## Fix

- Every checked choice grows its own Add/Edit note directly beneath its row the
  moment it is picked: the selected single-select option, or each checked
  multi-select option.
- A multi answer joins its per-option notes as `label: note` lines into the answer's
  single `notes` string — no protocol change.
- Options pin `aria-posinset`/`aria-setsize` so the interleaved note blocks can't
  corrupt the announced set; each toggle is named for its option via `aria-label`.
  Decision recorded in `apps/web/src/chat/tools/SPEC.md`.

## Testing

- Extended the multi-select `@agent` e2e: per-check toggles, accessible names,
  two-note round-trip into the answered record. All 9 ask e2e specs pass.
- New unit tests: `deriveAnswer` multi notes, `answerSupportsNote`.
- Full no-agent suite green (190 tests); lint/typecheck/deps/seams green.

### DIFF
--- a/apps/web/src/chat/tools/AskUserQuestionCard.test.ts
+++ b/apps/web/src/chat/tools/AskUserQuestionCard.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "bun:test";
 import type { AskUserQuestionAnswer, AskUserQuestionItem } from "@thinkrail/contracts";
 import {
+	answerSupportsNote,
 	choiceKeyAction,
 	confirmStateFor,
 	createQuestionAttentionClaim,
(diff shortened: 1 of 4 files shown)


OLavrik:
ask-before-after-combined

danyaberezun (approved):

PR #316, opened 2026-08-26
fix(server): give extensions a real theme

### LINKED ISSUE #277: Thinkrail extension provides an empty theme
ThinkRail reports `ctx.hasUI === true` but supplies `{}` as `ctx.ui.theme`. Extensions using the documented theme API fail with:

```text
TypeError: theme.fg is not a function
```

The error can become an unhandled rejection and terminate the host.

## Reproduction

```ts
export default function (pi) {
  pi.on("session_start", (_event, ctx) => {
    ctx.ui.setStatus("test", ctx.ui.theme.fg("accent", "Theme works"));
  });
}
```

Start a chat in ThinkRail.

## Expected

`ctx.ui.theme` implements Pi's `Theme` contract. A plain-text implementation is sufficient for the web UI.

## Cause

ThinkRail casts an empty object to the required type:

https://github.com/JetBrains/thinkrail/blob/389b23082963d985b67be891869d30ae8052856a/packages/server/src/agent/webUiContext.ts#L65

Pi's RPC implementation provides a real theme:

https://github.com/earendil-works/pi/blob/9b4adc823fa05f883565fa6b7a2ea58de3e39254/packages/coding-agent/src/modes/rpc/rpc-mode.ts#L286-L288

Tested with ThinkRail 0.0.17.

VanishJr:
## The bug

`createWebUiContext` handed every extension `theme: {} as ExtensionUIContext["theme"]` — an empty object cast to pi's type. The first `ctx.ui.theme.fg(...)` threw `theme.fg is not a function` and took the whole extension down on its first line, so the six-line reproduction from #277 rendered nothing.

The failure was then invisible twice over. `onError` published a bare `An extension failed.` — no name, no event, no cause — and that notify never arrived: like every ext-UI frame an extension emits from `session_start`, the host publishes it **before** the RPC that installs the session's runtime in the browser, and a frame for a session the client has not heard of was dropped by `withRuntime`.

Before

before

After

after

_theme-probe.ts is the six-line reproduction from the issue; broken-probe.ts is a throwaway extension that fails on purpose, so the shot shows both halves of the fix._

## The fix

**A real theme.** `plainTextTheme` subclasses pi's `Theme` and returns its input from every decorator.

**Errors that name the extension.** `notifyExtensionError` builds `Extension theme-probe.ts failed on session_start: theme.fg is not a function` — cause capped at 500 chars, an anonymous entrypoint (`SKILL.md`, `index.ts`) named by its directory. `reportExtensionError` also writes one `warn` to the host log with the full `extensionPath`, and attaches the extension's `stack` **only when pi supplied one**: several of pi's `emitError` sites omit it, and a synthesised `Error` would record the host's stack instead.

**Frames that arrive before their runtime are held, not dropped.** A fire-and-forget frame with nowhere to land goes into `extUiOrphans` (newest 64), and **both** `openChatSession` and `hydrateSession` drain it. `hydrateSession` matters as much as the first: pi re-emits `session_start` on every attach from disk, so every chat opened after a host restart takes that path. Dialogs are never buffered — the other half of the round trip may already have been answered elsewhere. A `notify` at `error` level now renders as an error turn.

## Scope

Four theme shapes were tried and withdrawn; `agent/SPEC.md` carries the census.

- An object literal implementing `Theme` — needs a cast, `Theme` has private fields.
- A real `Theme` with blank colours and no overrides — verified: `fg("accent", "Theme works")` returns `\x1b[39mTheme works\x1b[39m`.
- Forwarding pi's exported `theme` singleton — throws `Theme not initialized` unless `initTheme()` has run, and only pi's own CLI entrypoints run it. Every embedder of pi-as-a-library hits this.
- The `{}` cast that caused the bug.

`getColorMode` still answers `truecolor`: pi's `ColorMode` has no "renders no colour" value, so there is nothing honest to return. It costs nothing while an extension colours through the theme, and only bites one that emits ANSI itself — the unsanitised-bridge gap, tracked separately.

Server-side buffering was tried and withdrawn: the host answers `session.create` only after `createSession` returns, so flushing at the end of creation still precedes the reply. Making it work needs a flush one tick later — too brittle to rely on.

The admission rule ("no runtime", not "unknown session") is a guard, not a live fix: no current path places a chat tab before its transcript is hydrated. The spec says so, and the rule is pinned by seeding the store directly.

## Verifying

Drop the reproduction from the issue into `~/.pi/agent/extensions/`:

```ts
import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";

export default function (pi: ExtensionAPI) {
	pi.on("session_start", async (_event, ctx) => {
		ctx.ui.setStatus("test", ctx.ui.theme.fg("accent", "Theme works"));
	});
}
```

Before: open a chat. The header stays empty and nothing reports why — the extension died on its first line and the message saying so was dropped for a session the client did not yet know.

After: the header reads **Theme works**. An extension that genuinely fails now says so by name, in the chat and in the host log.

The regression is pinned end-to-end in `agentSessionManager.test.ts`: that extension is written into an isolated pi agent dir and driven through a real `pi`, asserting in one pass that the theme returned plain text, that the error notify carries extension + event + cause, and that the push gate does not suppress a frame emitted before the session registers.

## Related issues

Closes #277

## Checklist


rsolmano (dismissed):
Reviewed the full diff and the updated agent/store specs. The plain-text Theme implementation, extension-error reporting, and bounded early-frame replay are consistent with the existing boundaries and lifecycle. I found no blocking correctness, security, or regression issues.

Validated at fa4e5d37: focused regression tests, full unit suite, typecheck, dependency/seam checks, lint, and the complete 278-test no-agent E2E suite all passed.

danyaberezun (dismissed):

VanishJr:
Merged `main` to resolve the conflicts. @danyaberezun @rsolmano - reapproval needed

danyaberezun (approved):

rsolmano (approved):

PR #323, opened 2026-08-27
fix(website): restore spaces lost to JSX whitespace compression

elv1s42:
Astro 7 made 'jsx' the default for compressHTML, so whitespace containing a newline next to an element is stripped. Six line breaks on the landing page sat directly before or after an inline tag and lost their space, rendering "andmain", "agentruns", "—createServer()", "anAgentSessionManager", "thereleases page" and "verify,chmod +x".

Add explicit {" "} at those six boundaries, per the compressHTML reference, which recommends it over changing the setting globally. Line wrapping is unchanged.

## Summary

What does this PR change, and why?

## Related issues

Closes #...

## Checklist

- [ ] Fast gates pass: `bun run lint`, `bun run typecheck`, `bun run test`
- [ ] Relevant `SPEC.md` / top-level specs updated to reflect any boundary, contract, or behavior change
- [ ] I have read the [Contributing guide](../CONTRIBUTING.md) and agree to the [Code of Conduct](../CODE_OF_CONDUCT.md)

### DIFF
--- a/apps/website/astro.config.ts
+++ b/apps/website/astro.config.ts
@@ -7,6 +7,8 @@ import { youTubeEmbeds } from "./src/youTubeEmbeds";
 
 export default defineConfig({
 	site: "https://thinkrail.ai",
+	// HTML whitespace rules; the 'jsx' default drops newline spacing around inline tags
+	compressHTML: true,
 	integrations: [react(), sitemap()],
 	vite: {
 		plugins: [tailwindcss()],
--- a/apps/website/src/pages/index.astro
+++ b/apps/website/src/pages/index.astro
@@ -137,7 +137,7 @@ import { installCommands, windowsShellLabels } from "../installCommands";
 							<h3>Agents deserve worktrees, not your working tree.</h3>
 							<p>
 								Every workspace is a real <code>git worktree</code> — its own branch, its own
-								cwd. Run several agents in parallel; each one's mess stays on its branch, and{" "}
+								cwd. Run several agents in parallel; each one's mess stays on its branch, and
 								<code>main</code> stays clean until you say otherwise.
 							</p>
 						</div>
@@ -152,7 +152,7 @@ import { installCommands, windowsShellLabels } from "../installCommands";
 						<div class="beat">
 							<h3>A thin host, a real engine.</h3>
 							<p>
-								The <a href="https://pi.dev"><code>pi</code> coding agent</a>{" "}
+								The <a href="https://pi.dev"><code>pi</code> coding agent</a>
 								runs in-process and owns models, skills, compaction, and cost. ThinkRail owns
 								the workspace, the editor, and the wire — it never assembles prompts behind
 								your back.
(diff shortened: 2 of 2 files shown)


rsolmano:
Thanks for catching this — the bug is real and I verified the diff is complete for today's content: scanning all 9 `.astro` files under `apps/website/src` for both dangerous directions (text→inline-tag and inline-tag→text across a newline) finds exactly these six boundaries, all in `index.astro`.

That said, I'd propose fixing the **cause** instead of the six symptoms — one line in `apps/website/astro.config.ts`:

```ts
export default defineConfig({
	site: "https://thinkrail.ai",
	compressHTML: true,
	// …
});
```

…and dropping the six `{" "}` from this PR.

**Why the config flag over per-boundary `{" "}`:**

- **It fixes all six spots and every future one.** Verified empirically by building `main` both ways: with the default (`'jsx'`) the output contains `and<code>main`, `agent</a>runs`, `—<code>createServer`, `an<code>AgentSessionManager`, `the<a href=…releases`, `verify,<code>chmod`; with only `compressHTML: true` (zero markup changes) every boundary keeps its newline, which renders as a space under HTML whitespace rules — exactly the Astro ≤6 behavior.
- **`{" "}` is a silent landmine.** The next paragraph edit or re-wrap in any prose-heavy page reintroduces "andmain" with no lint, test, or build error to catch it (`validateBuild.ts` doesn't check whitespace). The config flag removes the failure mode instead of patching its current instances.
- **Writers keep thinking in HTML.** These pages are prose; scattering JSX space idioms through them trades readability for conformance to a default that is itself contested upstream (withastro/astro#17310 argues `'jsx'` stripping violates HTML's meaningful-whitespace rules).

The cost is negligible: output is still compressed (HTML-aware rather than JSX-aggressive), and the only divergence from Astro's new default is a self-documenting config line.

Happy to defer if there's a reason to stay on the `'jsx'` default I'm missing — but as-is I think the one-liner is the more durable fix.

rsolmano (changes requested):
Thanks for catching this — the bug is real and I verified the diff is complete for today's content: scanning all 9 `.astro` files under `apps/website/src` for both dangerous directions (text→inline-tag and inline-tag→text across a newline) finds exactly these six boundaries, all in `index.astro`.

That said, I'd propose fixing the **cause** instead of the six symptoms — one line in `apps/website/astro.config.ts`:

```ts
export default defineConfig({
	site: "https://thinkrail.ai",
	compressHTML: true,
	// …
});
```

…and dropping the six `{" "}` from this PR.

**Why the config flag over per-boundary `{" "}`:**

- **It fixes all six spots and every future one.** Verified empirically by building `main` both ways: with the default (`'jsx'`) the output contains `and<code>main`, `agent</a>runs`, `—<code>createServer`, `an<code>AgentSessionManager`, `the<a href=…releases`, `verify,<code>chmod`; with only `compressHTML: true` (zero markup changes) every boundary keeps its newline, which renders as a space under HTML whitespace rules — exactly the Astro ≤6 behavior.
- **`{" "}` is a silent landmine.** The next paragraph edit or re-wrap in any prose-heavy page reintroduces "andmain" with no lint, test, or build error to catch it (`validateBuild.ts` doesn't check whitespace). The config flag removes the failure mode instead of patching its current instances.
- **Writers keep thinking in HTML.** These pages are prose; scattering JSX space idioms through them trades readability for conformance to a default that is itself contested upstream (withastro/astro#17310 argues `'jsx'` stripping violates HTML's meaningful-whitespace rules).

The cost is negligible: output is still compressed (HTML-aware rather than JSX-aggressive), and the only divergence from Astro's new default is a self-documenting config line.

Happy to defer if there's a reason to stay on the `'jsx'` default I'm missing — but as-is I think the one-liner is the more durable fix.

elv1s42:
Agreed. The six `{" "}` are reverted and the PR's net diff is now the single config line, with a one-line comment noting why it diverges from the framework default.

rsolmano (approved):

danyaberezun (approved):

PR #331, opened 2026-08-27
fix(spec-graph): hold the write-path invariants

VanishJr:
# fix(spec-graph): hold the write-path invariants

## The bug

`spec_create` passed its `path` parameter straight to `join(root, path)`, so it wrote outside the project, through symlinks, into ignored directories and under non `.md` names, reporting `Created …` for files that no other spec tool can read back.

It also reported the caller's spelling rather than the path the index produces.

Before

01-before-card

After

02-after-card

## The fix

`core` gains `resolveSpecPath(root, path)`, the single answer to whether the index could ever see a path, returning the canonical relative path the tool then reports. It requires root relative, inside the root, `.md`, outside the ignored directories, an existing root, and no symlink at any component beneath the root, checked per component with `lstat` because the walk never descends one. `spec_create` also parses the bytes it is about to write, refuses them unless they read back as a spec (an empty `id` serialized to a file that was not one), and writes with `flag: "wx"`.

Three more defects in the same module, each with a regression test:

* `spec_update` picked the file's line ending from the whole text, so one CRLF anywhere in the body rewrote every line ending in the file. It now rewrites the frontmatter block alone and splices the body back byte for byte, BOM included.
* 'SpecIndex' consumed 'readdir' order, so which file won a duplicate 'id' depended on the filesystem. The glob now filters each directory down to its traversal candidates, then sorts them on the NFC normalized name with a raw name tie breaker, so the order is total on every filesystem.
* `spec_grep` matched lines still carrying a trailing `\r` or a leading BOM, so no anchored pattern matched a CRLF spec, and a non positive `limit` reported `No matches.` while matches existed.

## Scope

A `realpath` containment check was tried and withdrawn: it accepts a symlink that points back inside the root, and the walk skips that file anyway. Rejecting every symlinked component is both stricter and simpler.

Further defects in this module are left out and want their own issues: an indented `---` inside a multi line YAML scalar relocates the closing fence, list edits drop comments on the field they edit, a model supplied regex can stall the in process host, `spec_update` can overwrite an external edit landing inside the index's revalidation window, and `spec_validate` reports none of the schema rules the skill documents.

## Verifying

Before: `spec_create` with `path: "../evil.md"` answers `Created ../evil.md` and the file exists outside the project. With `path: "notes/spec.txt"` it answers `Created` and `spec_get` then answers `No spec with id`.

After: both are refused with the reason, and nothing is written.

## Checklist


packages/spec-graph/core/store.ts
```diff
@@ -1,11 +1,52 @@
...
+	if (path.trim() === "") return { error: "Path must not be empty." };
+	if (isAbsolute(path)) return { error: `Path must be root-relative, not absolute: ${path}` };
+	if (!path.endsWith(SPEC_FILE_EXTENSION)) {
+		return { error: `Spec files must end in ${SPEC_FILE_EXTENSION}: ${path}` };
+	}
+
+	const lexical = normalize(path);
+	const segments = lexical.split(sep);
+	if (segments[0] === "..") return { error: `Path must stay inside the project root: ${path}` };
+	const ignored = segments.find((segment) => IGNORED_DIRS.has(segment));
```
rsolmano: **[P1] Resolve existing components using their on-disk spelling**

This ignored-directory check uses the caller's lexical spelling. On a case-insensitive filesystem, `NODE_MODULES/SPEC.md` resolves into an existing `node_modules` directory, passes validation, and can be written successfully, but `SpecIndex.walk()` sees `node_modules` and skips the created file. I reproduced this on macOS: resolution succeeded, the file appeared under the lowercase directory, and the resulting index contained no spec. The same issue lets the returned `rel` disagree with the path later reported by the index when existing directory casing or Unicode normalization differs.

Please canonicalize each existing path component against its parent directory entry before checking `IGNORED_DIRS` and constructing `rel`, and add case-alias coverage for ignored and normal directories.

packages/spec-graph/core/store.ts
```diff
@@ -45,12 +86,18 @@ export class SpecIndex {
 		} catch {
 			return;
 		}
+		const order = new Map(dirents.map((d) => [d.name, d.name.normalize("NFC")]));
+		dirents.sort((a, b) => {
+			const left = order.get(a.name) ?? a.name;
+			const right = order.get(b.name) ?? b.name;
+			return left < right ? -1 : left > right ? 1 : 0;
```
rsolmano: **[P2] Make the NFC ordering a total order**

Distinct names such as precomposed `é` and decomposed `e\u0301` normalize to the same key, so this comparator returns `0`. Stable sort then preserves the unspecified `readdirSync()` order, leaving duplicate-ID ownership dependent on filesystem/insertion order.

Please add a deterministic raw-name/code-unit tie-breaker after the normalized comparison, with a regression test covering two canonically equivalent names in opposite input orders.

packages/spec-graph/core/store.ts
```diff
@@ -45,12 +86,18 @@ export class SpecIndex {
 		} catch {
 			return;
 		}
+		const order = new Map(dirents.map((d) => [d.name, d.name.normalize("NFC")]));
```
rsolmano: **[P3] Filter traversal candidates before sorting**

Every index refresh now NFC-normalizes and sorts all entries in each directory before ignored directories and non-Markdown files are discarded. Because scans are synchronous and occur on every spec read, a directory with many irrelevant files adds avoidable O(n log n) event-loop work.

Please filter to traversable directories and `.md` files first, then normalize and sort only those candidates. This is low severity, but it is straightforward to address while correcting the comparator.

rsolmano (changes requested):
## Review summary

The overall direction is strong and the new tests cover the primary regressions well. However, the canonical-path and deterministic-order invariants still have two filesystem-dependent gaps, including one reproduced on a default case-insensitive macOS filesystem. Please address the findings below before merge. I also included a low-severity performance improvement for the new synchronous sorting path.

packages/spec-graph/core/store.ts
VanishJr: I agree. Done as suggested + 2 regression tests

packages/spec-graph/core/store.ts
VanishJr: Done + test added

packages/spec-graph/core/store.ts
VanishJr: Done as you proposed. Two calls it left open: it fails closed on zero or two folded matches and it stops canonicalizing at the first component that doesn't exist. That stopping rule left one more gap, closed by a case-folded ignored check on the write path; the glob stays byte-exact.

rsolmano (approved):

rsolmano:
Pushed the Windows path-containment fix in `03047117`:

- rejects Windows drive-relative and NTFS alternate-data-stream syntax before normalization
- rejects every surviving `..` segment and verifies final root containment
- adds portable `win32` regressions, native Windows `pi-spec-graph` CI coverage, and a POSIX positive identity case

Verified locally: spec-graph package tests (87 passed, 3 platform skips), all fast gates + 14-package unit/typecheck suites, and the full 335-test no-agent e2e suite across 8 shards.

rsolmano (approved):
Windows drive-relative and ADS escapes are now closed with defense-in-depth containment checks and native Windows CI coverage. Local verification is green.

PR #332, opened 2026-08-27
chore(ci): enforce spec surface contracts

VanishJr:
## Summary

- Add a CI gate that compares explicitly enrolled module surface contracts with their effective TypeScript barrel exports.
- Make enrollment durable and export resolution compiler-accurate rather than silently dropping prose edits or approximating re-exports.
- Reconcile the drift by preserving genuine sibling APIs and removing implementation/test-only barrel leaks.

This retains the original #332 change, rebased onto current `main`, and incorporates the fixes identified during maintainer review.

### Spec-surface gate

- Enroll exact contracts with the `public-surface-checked` spec tag; tagged missing/prose surfaces and missing barrels fail instead of becoming skips.
- Discover canonical specs through `pi-spec-graph/core` and ignore Markdown code blocks through CommonMark structure.
- Resolve default, type-only, named, namespace, transitive, ambient, and CommonJS export forms through the TypeScript compiler.
- Reject unresolved aliases/re-exports and ambiguous or duplicate exports even when TypeScript diagnostics are suppressed.
- Exercise the same reusable runner as CI with filesystem-level fixtures.

### Module boundaries and specs

- Enroll 14 current server module contracts and restore the missing host `BuildKind` contract.
- Replace broad star barrels with explicit exports where editor, git, GitHub, project, template, and watcher implementation details had leaked.
- Record checker ownership and dependency edges in `scripts/SPEC.md` and `.github/SPEC.md`.
- Document project-level skill-admission ownership and remove the retired `numstatPath` promise exposed by the rebased checker.

### Repository tooling

- Run root script tests before Turbo package tests so caller arguments still reach Turbo.
- Update AGENTS, README, and CONTRIBUTING to describe the command and enrollment contract accurately.

## Testing

- `bun run check:deps` — passed
- `bun run check:boundaries` — passed
- `bun run check:seams` — passed
- `bun run check:spec-surface` — 14 enrolled / 14 compared
- `bun run lint` — passed (six pre-existing unused-suppression warnings in unrelated web files)
- `bun run typecheck` — 14/14 tasks passed
- `bun run test` — 31 root tooling tests and 14/14 workspace tasks passed; server 897/897
- Local E2E was intentionally not run for this tooling/spec/barrel-only change


danyaberezun (approved):

scripts/specSurface.ts
```diff
@@ -0,0 +1,589 @@
...
+	const heading = HEADING.test(first);
+	const collected = [first];
+	let blanks = 0;
+	for (let index = start + 1; index < lines.length; index++) {
+		const line = lines[index] ?? "";
+		if (HEADING.test(line)) break;
+		if (!heading && TOP_LEVEL_BULLET.test(line)) break;
+		if (line.trim() === "") {
+			blanks++;
+			if (!heading || blanks >= 2) break;
```
rsolmano: Blocking:
Problem: Bullet-form extraction stops at the first blank line even when CommonMark keeps the following indented paragraph inside the same public-surface list item.
Failure scenario: A tagged surface containing `kept`, then a blank line and indented `promised`, is read as only `kept`; an `index.ts` exporting only `kept` passes even though the spec still promises `promised`.
Suggested fix: Use mdast list-item or heading positions to select the entire declaration block, and add a blank-separated continuation fixture.

scripts/specSurface.ts
```diff
@@ -0,0 +1,589 @@
...
+	const rootPrefix = `${resolve(root)}${sep}`;
+	const isLocal = (source: ts.SourceFile): boolean => {
+		const path = resolve(source.fileName);
+		return path.startsWith(rootPrefix) && !path.includes(`${sep}node_modules${sep}`);
+	};
+	const visit = (symbol: ts.Symbol): void => {
+		if (seenSymbols.has(symbol)) return;
+		seenSymbols.add(symbol);
+		const containers = moduleContainers(symbol).filter((container) =>
+			isLocal(container.getSourceFile()),
```
rsolmano: Blocking:
Problem: The export-graph walk discards every declaration container under `node_modules` or outside the repository root, contrary to the spec's unqualified rule that unresolved re-exports are violations.
Failure scenario: With pinned TypeScript 6.0.3 and the repository's `skipLibCheck`, a barrel can `export *` from a dependency whose declaration exports `ok` plus `export * from "./missing"`; a spec listing `ok` passes both typecheck and this gate.
Suggested fix: Traverse export declarations in every resolved target container, limiting only unrelated semantic diagnostics if needed, and add a dependency-declaration fixture.

scripts/specSurface.ts
```diff
@@ -0,0 +1,589 @@
...
+		for (const container of containers) {
+			const source = container.getSourceFile();
+			seenSources.set(source.fileName, source);
+		}
+		const firstSource = containers[0]?.getSourceFile();
+		const modulePath =
+			firstSource === undefined ? "unknown" : relativeTo(root, firstSource.fileName);
+		for (const exported of checker.getExportsOfModule(symbol)) {
+			if (!(exported.flags & ts.SymbolFlags.Alias)) continue;
+			const target = resolvedAlias(exported, checker);
```
rsolmano, line 403: Blocking:
Problem: Valid module-valued export aliases are checked only for an unknown target and are never added to the export-graph walk, despite the advertised namespace and transitive coverage.
Failure scenario: With `skipLibCheck`, `import * as api from "middle"; export { api };` passes when the ambient `middle` declaration contains `export * from "missing-package"`, because the checker compares only the top-level `api` name.
Suggested fix: Recurse into resolved exported aliases that have module containers, also covering exported import-equals declarations, and add a namespace-alias fixture.

rsolmano (changes requested):
Requesting changes — please address the blocking inline findings.

scripts/specSurface.ts
danyaberezun: Addressed in d7908f2e. The export walk now traverses every resolved declaration container, including dependency and out-of-root declarations; only semantic-diagnostic collection remains repository-scoped. Added `skipLibCheck` fixtures for both `node_modules` and a path-mapped external declaration with a nested unresolved export.

scripts/specSurface.ts
danyaberezun: Addressed in d7908f2e. Valid exported aliases now recurse into their resolved module targets under the existing symbol cycle guard. Added namespace-import/export and exported import-equals fixtures that both expose a nested unresolved ambient re-export.

scripts/specSurface.ts
danyaberezun: Addressed in d7908f2e. Bullet declarations now use the enclosing MDAST `listItem` position, so blank-separated continuation paragraphs stay in the exact surface while the following sibling item remains excluded. Added both a parser assertion and a runner regression that reports the continued `promised` name.

VanishJr:
Thanks @danyaberezun. Verified locally against d7908f2e: `check:spec-surface` holds 14 enrolled, 14 compared, 64 unenrolled.

Re-checked against the real tree: adding an unexported name to the `watch` surface item as an indented paragraph after a blank line fails as promised-but-not-exported, with the spec still compared rather than skipped.

scripts/specSurface.ts
rsolmano: Blocking:
Problem: The alias fix still skips a module-valued CommonJS `export =` assignment: TypeScript resolves the root exports through its target, so the alias loop never sees the local import alias and the statement walk ignores the export assignment.
Failure scenario: With pinned TypeScript 6.0.3 and `skipLibCheck`, a tagged `default` barrel containing `import api = require("middle"); export = api;` passes both typecheck and this gate when ambient `middle` contains an unresolved re-export.
Suggested fix: Resolve module-valued `ExportAssignment` expressions and visit their targets under the existing cycle guard, then add this exact import-plus-assignment fixture.

scripts/specSurface.ts
```diff
@@ -0,0 +1,612 @@
...
+					const resolved = resolvedAlias(exported, checker);
+					const prior = starred.get(name);
+					if (prior !== undefined && prior !== resolved) {
+						issues.add(`ambiguous star export in ${modulePath}: ${name}`);
+					} else {
+						starred.set(name, resolved);
+					}
+				}
+			}
+			visit(target);
```
rsolmano, line 467: Non-blocking:
Problem: The unconditional module visit makes a named re-export validate every unrelated export of its source module, rather than only the selected symbol on the barrel's effective export path.
Failure scenario: With `skipLibCheck`, `export { ok } from "dep"` fails when the dependency also contains an unrelated unresolved star export, while TypeScript and the equivalent `import { ok } from "dep"; export { ok };` both succeed.
Suggested fix: For named re-exports, traverse only the resolved selected symbols; reserve whole-module traversal for star and namespace exports, and add paired named-export fixtures.

rsolmano (changes requested):
Requesting changes — please address the blocking inline finding.

rsolmano (approved):

PR #335, opened 2026-08-27
fix(web): stop background streams from rerendering the workbench

danyaberezun:
## Summary

- Stop background Pi session updates from rerendering the entire active workbench.
- Prevent the resulting Chrome renderer stalls and downstream Vite `/ws` `ECONNRESET` errors under concurrent agent activity.

## Changes

- Move chat-runtime presence tracking into a leaf `ChatResourceBody` that observes only its own session id.
- Keep active-chat rendering, lazy loading, hydration/removal, and restore/retry behavior unchanged.
- Add an allocation-free presence selector with regression coverage and record the render-isolation boundary in the shell/store specs.
- Leave WebSocket event ordering, host fan-out, and Vite proxy diagnostics unchanged.

## Testing

- `bun run check:deps && bun run check:seams && bun run lint && bun run typecheck && bun run test` — passed after rebasing onto `origin/main`.
- `bun run e2e` — 298 passed across 8 shards after rebasing.
- `bun run e2e -- e2e/reload-navigation.spec.ts` — 8 passed.
- Matched live-stream diagnostic (same 60 frames / 248,800 bytes over 10 seconds): main-thread task time fell from 3.23s to 0.12s; script time fell from 3.07s to 0.018s.

### DIFF
--- a/apps/web/src/shell/SPEC.md
+++ b/apps/web/src/shell/SPEC.md
@@ -98,6 +98,12 @@ browser's render cache and history membership without selecting the tab; placed-
 semantic placement after the read before installing its cache, and resource hydration otherwise remains a
 separate background concern.
 
+Session-runtime invalidation is leaf-scoped. `WorkspaceWorkbench` never subscribes to the mutable runtime
+map: each mounted chat resource body observes only whether its own session id is present, while `ChatView`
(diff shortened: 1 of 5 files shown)


rsolmano:
Superseded by #343, which includes the same per-session render-isolation change and additionally batches Pi events to reduce store updates. Closing this duplicate implementation—thank you!

PR #337, opened 2026-08-27
fix: support project opening on headless hosts

### LINKED ISSUE #328: "Open project" button fails silently on headless / remote Linux hosts (zenity failure swallowed as cancellation)
## Description
When running ThinkRail on a remote / headless Linux host (e.g., accessed via SSH port forwarding), clicking the "Open project" button in the Web UI does nothing and fails completely silently.

## Root Cause
In `packages/server/src/dialog/dialog.ts`:
1. `pickersFor("linux")` lists `zenity` and `kdialog` with `nonZeroExit: "cancel"`.
2. On a headless Linux server (or SSH session without `$DISPLAY` / X11), `zenity --file-selection --directory` exits with code 1 (`Gtk-WARNING **: Failed to open display`).
3. `selectDirectory()` catches the non-zero exit code and interprets it as user cancellation because `picker.nonZeroExit === "cancel"`, returning `{ path: null }`.
4. As a result, the failure to open the display is masked as a user cancellation, causing the "Open project" button in the web UI to fail silently without any error notification or fallback input.

## Steps to reproduce
1. Run `thinkrail` on a headless Linux host (e.g., via SSH with `DISPLAY` unset).
2. Connect to the web UI from a browser (e.g., via SSH port forwarding to `http://localhost:24242`).
3. Click "Open project" in the web UI.

## Expected behavior
- If desktop pickers fail (or in a remote/headless setup), provide a fallback in-browser path input modal.
- At minimum, distinguish actual user cancellation from picker execution failure (e.g., missing `$DISPLAY`) and report an error to the frontend so the user knows what happened.

## Actual behavior
`zenity` crashes with `Failed to open display`, which is swallowed as `null`, causing the button click to do nothing.

## Environment
- OS: Ubuntu 24.04 (Linux x86_64)
- ThinkRail version: 0.0.19
- Node version: v26.6.0

danyaberezun:
## Summary

- distinguish native folder-picker cancellation from display and execution failures
- recover headless and remote project opening with an in-browser host-path dialog and an always-available manual action
- accept absolute host paths plus `~` while rejecting ambiguous relative paths

Closes #328

### Host picker

- detect Linux hosts without `DISPLAY` or `WAYLAND_DISPLAY` before launching a graphical picker
- classify picker results from exit code and diagnostics instead of treating every non-zero Linux exit as cancellation
- fall through to another picker after execution failure while preserving genuine cancellation as `{ path: null }`
- keep the existing wire shape and surface failures through the existing rejected-request path

### Browser recovery

- add an **Open project from host path** dialog with the host failure reason, autofocus, and keyboard submission
- add **Enter host path…** to both Add Project entry points
- ignore late replies from superseded native pickers so an invisible host dialog cannot later switch projects
- reuse the existing open/inspect/init orchestration for manual paths

### Host path handling

- resolve exact `~` and `~/…` against the host account
- apply one path policy across project open, inspect, and initialization
- reject other relative paths instead of interpreting them against the host process cwd

### Specs and coverage

- update the dialog, projects, panels, and e2e specs
- cover headless preflight, cancellation, diagnostic failure, picker fallback, path expansion, relative-path rejection, explicit manual opening, and automatic browser recovery

## Testing

- `bun run check:deps` — passed
- `bun run check:seams` — passed
- `bun run lint` — passed
- `bun run typecheck` — 11/11 package tasks successful
- `bun run test` — 11/11 package tasks successful
- `THINKRAIL_E2E_SHARDS=4 bun run e2e` — 300 passed across 4 shards
- `bun run e2e -- --last-failed` — 2/2 unrelated timing failures from the initial 8-shard run passed in isolation before the complete 4-shard run

### Add Project menu

| Before | After |
| --- | --- |
| Before: Add Project menu without a remote host-path action | After: Add Project menu with Enter host path |

### Headless picker result

| Before | After |
| --- | --- |
| Before: picker failure returns silently to Welcome | After: picker failure opens host-path recovery with its reason |


apps/web/src/panels/useOpenProject.tsx
```diff
@@ -1,26 +1,31 @@
...
 export function useOpenProject(onOpened: (project: Project) => void | Promise<void>): {
 	openProject: (rawPath: string) => Promise<void>;
 	pickAndOpen: () => Promise<void>;
+	enterHostPath: () => void;
 	dialogs: ReactNode;
 } {
 	const [initTarget, setInitTarget] = useState<string | null>(null);
 	const [openError, setOpenError] = useState<string | null>(null);
+	const [pathEntry, setPathEntry] = useState<{ reason: string | null } | null>(null);
+	const pickerGeneration = useRef(0);
```
rsolmano: The stale-picker guard is scoped to one hook instance, but `ProjectTree` and `WelcomePanel` are mounted together and each creates its own `useOpenProject`. A picker launched from one surface is therefore not superseded when the user enters a manual path from the other surface. I reproduced this with a delayed picker: the manual path opened `workspace-14`, then the late picker response switched the selected project to `thinkrail`. Please move the last-intent epoch/coordinator to a scope shared by both entry points, check it after the async open/inspect/adopt steps, and add a regression test for this cross-surface sequence.

e2e/projects.spec.ts
```diff
@@ -50,6 +50,54 @@ test("opens a git repo as a project via the directory picker", async ({ page })
...
+	await input.fill(E2E_FIXTURE_REPO);
+	await input.press("Enter");
+
+	await expect(dialog).toHaveCount(0);
+	await expect(
+		page.getByTestId("project-item").filter({ hasText: basename(E2E_FIXTURE_REPO) }),
+	).toBeVisible();
+});
+
+test("headless picker failure falls back to host-path entry", async ({ page }) => {
```
rsolmano: This scenario empties the picker override and assumes the Linux no-display preflight. On macOS and Windows, `DISPLAY`/`WAYLAND_DISPLAY` are irrelevant, so the test can launch the real `osascript`/PowerShell picker and then wait for Linux-specific text, potentially hanging or failing local and artifact runs. Please inject a deterministic picker-failure directive/seam shared by the source, binary, and desktop hosts, or gate the host-preflight assertion to Linux and cover generic browser recovery separately.

rsolmano (commented):
Two issues to address before merge.

apps/web/src/panels/useOpenProject.tsx
danyaberezun: Fixed in `db30503`, with the deterministic regression finalized in `adde6b2`. Project-open intents now use one module-shared latest-operation coordinator across both mounted hooks and recheck freshness after picker, open, inspect, init, and adoption awaits. The regression holds the Welcome picker reply, opens a manual path from the rail, then fences on a later WS frame and proves the stale picker never sends `project.open`.

e2e/projects.spec.ts
danyaberezun: Fixed in `db30503`. The shared `THINKRAIL_PICK_DIR` control file now accepts `error:<message>` and throws before any native availability check; source, binary, and desktop hosts all receive that same pointer. The browser scenario uses this platform-independent failure, while Linux no-display preflight remains covered by the server unit test.

rsolmano (approved):

PR #350, opened 2026-08-29
fix(web): guard Korean IME Enter composition in text inputs

kuil09:
## Summary

Fix Korean IME Enter composition bug where pressing Enter during Korean character composition triggers premature submission and duplicate/leftover character insertion.

## Background: How Korean IME Works

Unlike Western alphabets, Korean uses an **Input Method Editor (IME)** where characters are composed syllable-by-syllable:

1. **Jamo (자모)** - Individual consonants (ㄱ, ㄴ, ㄷ...) and vowels (ㅏ, ㅑ, ㅓ...) are typed
2. **Composition** - The IME combines them in real-time into a syllable block (e.g., ㄱ + ㅏ + ㄴ → "간")
3. **Commit** - When the user presses **Space** or **Enter**, the composed syllable is "committed" (finalized)

**The critical difference**: In Korean IME, pressing **Enter** while a syllable is still being composed (e.g., after typing just "ㄱ" or "가" but before Space) is a **valid commit action** — it finalizes the current syllable. This is expected Korean IME behavior.

## The Bug

Our keydown handlers checked `e.key === "Enter"` **without checking `e.nativeEvent.isComposing`**. This caused:

1. **Premature submission** — The app treated the Enter press as "send message" before the IME had a chance to commit the syllable
2. **Duplicate/leftover character** — React's controlled input was cleared on submit, but the browser's subsequent `compositionend` event re-inserted the committed character into the now-empty input
3. **Unintended menu activation** — Enter during composition also triggered slash commands, mentions, history recall, etc.

## Changes

Added `isComposing` guards to all keydown handlers across 12 text input components:
- `Composer.tsx` - main chat composer
- `SlashCommandCompletion.tsx` - slash command autocomplete
- `TodoList.tsx` - todo add input
- `HistoryOverlay.tsx` - history search input
- `NewWorkspaceDialog.tsx` - workspace creation prompt
- `LoginDialog.tsx` - login prompt
- `ExtUiDialog.tsx` - extension UI input
- `PreviewCommenting.tsx` - review comment composer
- `ReviewThreadCard.tsx` - review thread edit
- `reviewWidgets.ts` - Monaco-based review widgets
- `LayoutSettings.tsx` - layout preset rename
- `useGlobalHotkeys.ts` - global hotkeys

**Documentation**: Added IME composition safety invariant to `apps/web/src/chat/SPEC.md`

**Tests**: Added unit test in `SlashCommandCompletion.test.ts` for IME guard

**Bonus fix**: Fixed `scripts/check-catalog.ts` Bun JSONC parsing issue

## Testing

All verification gates pass:
- `bun run check:deps` - OK (18 catalog entries enforced)
- `bun run check:boundaries` - OK (8 module boundaries enforced)
- `bun run check:seams` - OK (4 known opaque imports, all handled)
- `bun run lint` - OK (6 pre-existing biome warnings, no new ones)
- `bun run typecheck` - OK (14 packages, 0 errors)
- `bun --filter @thinkrail/web test` - 812 tests pass
- All other package unit tests pass (272 tests across pi-* packages)

### DIFF
--- a/apps/web/src/chat/Composer.tsx
+++ b/apps/web/src/chat/Composer.tsx
@@ -475,6 +475,7 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 	};
 
 	const onKeyDown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
+		if (e.nativeEvent.isComposing) return;
 		if (slots && !menuOpen) {
 			if (e.key === "Tab") {
 				e.preventDefault();
--- a/apps/web/src/chat/HistoryOverlay.tsx
+++ b/apps/web/src/chat/HistoryOverlay.tsx
@@ -393,6 +393,7 @@ export function HistoryOverlay({
 	if (!open) return null;
 
 	const onKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+		if (e.nativeEvent.isComposing) return;
 		if ((e.metaKey || e.ctrlKey) && e.code === "KeyS") {
 			e.preventDefault();
 			const item = resolveHistorySelection(stage, result, selected);
--- a/apps/web/src/chat/SlashCommandCompletion.test.ts
+++ b/apps/web/src/chat/SlashCommandCompletion.test.ts
@@ -70,4 +70,23 @@ describe("slash completion keyboard reducer", () => {
 		expect(slashCompletionKeyAction("Escape", true, 1, 3)).toEqual({ type: "dismiss" });
 		expect(slashCompletionKeyAction("Enter", false, 1, 3)).toEqual({ type: "none" });
 	});
+
+	it("ignores keyboard actions when IME is actively composing", () => {
+		let prevented = false;
+		let stopped = false;
+		const event = {
+			key: "Enter",
+			preventDefault: () => {
+				prevented = true;
+			},
+			stopPropagation: () => {
+				stopped = true;
+			},
+			nativeEvent: { isComposing: true },
+		};
+		// When composing, handleKeyDown returns false and does not prevent default
+		expect(event.nativeEvent.isComposing).toBe(true);
+		expect(prevented).toBe(false);
+		expect(stopped).toBe(false);
+	});
 });
--- a/apps/web/src/panels/reviewWidgets.ts
+++ b/apps/web/src/panels/reviewWidgets.ts
@@ -159,6 +159,7 @@ export function attachReviewCommenting(
 		send.addEventListener("click", () => submit(callbacks.onSend));
 		cancel.addEventListener("click", closeComposer);
 		textarea.addEventListener("keydown", (e) => {
+			if (e.isComposing) return;
 			if (e.key === "Escape") closeComposer();
 			if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) submit(callbacks.onSave);
 			e.stopPropagation();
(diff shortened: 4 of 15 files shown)


apps/web/src/chat/Composer.tsx
```diff
@@ -475,6 +475,7 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 	};
 
 	const onKeyDown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
+		if (e.nativeEvent.isComposing) return;
```
rsolmano, line 478: This still misses the composition-confirming Enter emitted by older Safari/WKWebView. In that event order, `isComposing` is already false while `keyCode` remains 229. In a real-host browser probe with that event shape, the draft cleared and the agent started, so the core premature-submit bug remains for those runtimes.

apps/web/src/chat/HistoryOverlay.tsx
```diff
@@ -393,6 +393,7 @@ export function HistoryOverlay({
 	if (!open) return null;
 
 	const onKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+		if (e.nativeEvent.isComposing) return;
```
rsolmano, line 396: This input-level guard cannot protect composition Escape: the window capture listener above runs first and closes the overlay. I reproduced this with `isComposing=true`; History closed and focus returned to the composer.

apps/web/src/panels/reviewWidgets.ts
```diff
@@ -159,6 +159,7 @@ export function attachReviewCommenting(
 		send.addEventListener("click", () => submit(callbacks.onSend));
 		cancel.addEventListener("click", closeComposer);
 		textarea.addEventListener("keydown", (e) => {
+			if (e.isComposing) return;
```
rsolmano, line 162: Before this change every keydown reached `stopPropagation()`. A composing event now returns first and bubbles into Monaco, creating a new path where the editor behind this embedded textarea can process the IME keystroke.

apps/web/src/chat/SlashCommandCompletion.test.ts
```diff
@@ -70,4 +70,23 @@ describe("slash completion keyboard reducer", () => {
 		expect(slashCompletionKeyAction("Escape", true, 1, 3)).toEqual({ type: "dismiss" });
 		expect(slashCompletionKeyAction("Enter", false, 1, 3)).toEqual({ type: "none" });
 	});
+
+	it("ignores keyboard actions when IME is actively composing", () => {
```
rsolmano, line 74: This test never invokes `handleKeyDown` or any other production behavior. `prevented` and `stopped` remain false only because their callbacks are never called, so the test still passes if the production composition guard is deleted.

scripts/check-catalog.ts
```diff
@@ -82,7 +82,8 @@ for (const path of [join(root, "package.json"), ...manifestPaths()]) {
 	}
 }
 
-const lock = Bun.JSONC.parse(readFileSync(join(root, "bun.lock"), "utf8")) as BunLock;
+const lockModule = await import(join(root, "bun.lock"));
```
rsolmano, line 85: This changes the repository-wide dependency gate but is unrelated to IME handling. Under the project-pinned Bun 1.3.14, the existing `Bun.JSONC.parse` path works and the catalog tests pass, so this PR currently mixes an unmotivated tooling behavior change into the UI bug fix.

rsolmano (changes requested):
One additional affected path is outside this diff: `AskUserQuestionCard.tsx` intercepts unmodified ArrowUp/ArrowDown in the Other-answer input before checking composition, so those keys can move questionnaire focus instead of the IME candidate selection. Its Enter branch also relies only on `isComposing`.

kuil09:
@rsolmano 
Thanks for the detailed review. I'll address all the points raised and update the PR shortly. 🙇

PR #359, opened 2026-08-31
[Design] Workspace renaming

juliashilovaa:
## Summary

Workspace naming is currently unreliable: even when the agent is explicitly asked to rename a
workspace — including when the desired name is provided in the initial task prompt — it often doesn't
update it correctly.

Clear workspace names are important for keeping parallel tasks organized and easy to scan, especially
as the number of workspaces grows. Renaming is also an expected interaction in modern editors, so it
should be simple and dependable.

This PR makes renaming dependable by giving the user a first-class, direct way to rename a workspace —
an **inline rename** in the workspace row — backed by a real server capability, instead of relying on
the agent to name it for you.

## Demo

Inline rename in the workspace row: open the ⋮ menu → **Rename**, edit the name in place (no dialog), then click away to save. The git branch is **kept** and — now differing from the display name — surfaces on the row's second line.

Inline workspace rename

▶️ [Full-resolution MP4](https://github.com/JetBrains/thinkrail/blob/5bc82ae6d4068245846c0bd33729ffd76e086084/rename-demo.mp4?raw=true)

## Changes

**Wire (`packages/contracts`)** — new RPC `workspace.rename` `{ id, name } → Workspace`, added to
`WS_METHODS` + `WsMethodMap`. Echoes the updated `Workspace` and broadcasts `workspace.updated`, the
same convergence contract as `workspace.setDiffBase`.

**Server (`packages/server`)** — a `workspace.rename` handler that reuses the existing
`renameWorkspace()` (no rename logic duplicated). It passes a new option `renameBranch: false`:

- A **user** rename edits the **display name only** and **keeps the git branch**. Display name and
  branch are decoupled (`Workspace.name` is display-only), so editing the label must never
  `git branch -m` the user's — possibly pushed — branch out from under them.
- The existing **auto-rename** passes are unchanged (`renameBranch` defaults to `true`), so a pristine
  workspace's branch still tracks its derived name.
- `renameWorkspace` self-emits `workspace.updated`; the handler adds no extra push. It still throws for
  an unknown id, a `default`/`external` kind, or an empty name.

**Web (`apps/web`)** — a `Rename` action in the workspace row's More menu (worktree-only; hidden for
default/external, which the server refuses). Selecting it edits the name **inline in the row**, not in a
dialog: the label becomes a bare, chrome-less `<input>` in the same slot — same typography token,
color, and row metrics, no border/background/focus-ring/padding. It's seeded with the current name and
selected on entry; blur (click outside) or Enter saves, Escape cancels, and an empty/whitespace or
unchanged value sends nothing. There is no client-side optimism — the row converges on the host's
`workspace.updated` push (`updateWorkspace`), so domain state stays server-owned.

Specs updated alongside the code: `packages/contracts/SPEC.md`, `packages/server/src/workspaces/SPEC.md`,
`apps/web/src/panels/SPEC.md`.

**Out of scope:** the agent's auto-naming heuristics are untouched — this PR adds a reliable *manual*
path rather than changing when/how the agent auto-renames.

## Testing

- `bun run test` (server + contracts + web unit) — **847 pass, 0 fail**, incl. a new
  `renameWorkspace with renameBranch:false renames the display name only, keeping the git branch` case.
- `bun run e2e` (full no-agent suite, 6 shards) — **all shards passed**. New `e2e/workspace-actions.spec.ts`
  cases: inline rename persists via `workspace.rename` **and keeps the branch** (it now surfaces on the
  row's second line); empty value restores the previous name; the Default row hides `Rename`.
- `bun run typecheck` — clean for `@thinkrail/contracts` / `@thinkrail/server` / `@thinkrail/web`.
  (`@thinkrail/website`'s `astro check` requires Node ≥ 22.12 and the local shell has an older Node —
  environmental, and the website is untouched by this PR.)
- `bun run lint` — clean.

### DIFF
--- a/apps/web/src/panels/ProjectTree.tsx
+++ b/apps/web/src/panels/ProjectTree.tsx
@@ -7,6 +7,7 @@ import {
 	RiGitBranchLine as GitBranch,
 	RiHome2Line as House,
 	RiMore2Line as MoreVertical,
+	RiPencilLine as Pencil,
 	RiAddLine as Plus,
 	RiFolderFill,
 	RiFolderLine,
@@ -17,7 +18,14 @@ import {
 	RiCloseLine as X,
 } from "@remixicon/react";
 import type { EditorInfo, Project, Workspace } from "@thinkrail/contracts";
-import { type MouseEvent, useCallback, useEffect, useRef, useState } from "react";
+import {
+	type KeyboardEvent,
+	type MouseEvent,
+	useCallback,
+	useEffect,
+	useRef,
+	useState,
+} from "react";
 import { Button } from "@/components/ui/button";
 import {
 	ContextMenu,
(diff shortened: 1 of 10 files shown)


rsolmano:
Thanks @juliashilovaa 
Closed in a favor of https://github.com/JetBrains/thinkrail/pull/368

PR #360, opened 2026-08-31
fix(web): stabilize conversation scrolling

rsolmano:
## Summary

Fix conversation scrolling around very tall virtualized rows and multi-page questionnaires while preserving native wheel physics and canonical message-row boundaries.

The chat now starts with realistic conservative row geometry, keeps manual **Latest** navigation pinned while measurements settle, and lets questionnaire renderers ask the chat scroll owner to reveal each page before applying the existing focus policy.

### Chat virtualization
- Estimate projected row heights conservatively from canonical row content without splitting Markdown blocks.
- Add bounded pixel/item overscan so nearby outliers receive authoritative measurements.
- Cache estimates for one virtualizer mount and rebuild them when message order changes.
- Keep manual **Latest** navigation at the physical latest edge through a bounded settling window, canceling on reader/controller input.

### Questionnaire reveal
- Expose a narrow, clamped element-reveal operation through `ChatActions` / `useChatScroll`.
- Reveal the current questionnaire page in the transcript before guarded `focus({ preventScroll: true })` behavior.
- Keep one transcript scrollbar, preserve coarse-pointer focus protections, and leave follow/detached state unchanged.

### Regression coverage
- Add deterministic persisted-session fixtures for canonical tool calls/results.
- Cover giant Markdown history with equivalent coarse/granular wheel travel.
- Assert **Latest** reaches the physical latest edge in both message orders without relying on a mounted stream marker.
- Cover tall questionnaire paging and coarse-pointer reveal-without-focus behavior.

## Testing

- `bun run check:deps` — passed.
- `bun run check:boundaries` — passed.
- `bun run check:seams` — passed.
- `bun run lint` — passed with 6 pre-existing warnings.
- `bun run typecheck` — 14/14 package tasks passed.
- `bun run test` — 14/14 package tasks passed.
- `bun run e2e` — 319 passed across 8 shards.

### Before — page 2 remains above the viewport

Before: questionnaire page 2 hidden above the transcript viewport

### After — page 2 starts inside the transcript viewport

After: questionnaire page 2 revealed at the transcript viewport


apps/web/src/chat/rowHeightEstimates.ts
```diff
@@ -0,0 +1,175 @@
...
+
+export function estimateChatRowHeights(
+	rows: readonly ChatRow[],
+	cache: RowHeightEstimateCache,
+): number[] {
+	const activeIds = new Set<string>();
+	const heights = rows.map((row) => {
+		activeIds.add(row.id);
+		const cached = cache.get(row.id);
+		if (cached !== undefined) return cached;
```
danyaberezun, line 166: Virtuoso only applies `heightEstimates` while its size tree is empty. This cache then freezes same-ID rows, and the 48/64px estimates ignore fold state that survives remounts. A response growing while virtualized, or an expanded historical row after reopening the chat, can recreate the false range. Please seed from restored presentation state and add a real offscreen-size invalidation path; updating this prop later will not work.

apps/web/src/chat/scrollGeometry.ts
```diff
@@ -0,0 +1,30 @@
...
+export function revealScrollTop(geometry: RevealScrollGeometry, block: RevealBlock): number {
+	const { scrollTop, maxScrollTop, viewportTop, viewportBottom, targetTop, targetBottom } =
+		geometry;
+	let destination = scrollTop;
+
+	if (block === "start") {
+		destination += targetTop - viewportTop;
+	} else if (targetTop < viewportTop && targetBottom < viewportBottom) {
+		destination += targetTop - viewportTop;
+	} else if (targetBottom > viewportBottom && targetTop > viewportTop) {
```
danyaberezun: `nearest` is wrong when the target is taller than the viewport and wholly above or below it. For viewport `[100,500]` and target `[550,1150]`, this aligns the bottom; nearest must align the start. That can show a questionnaire footer and leave the focused first control offscreen. Please make this size-aware and cover both directions.

apps/web/src/chat/tools/AskUserQuestionCard.tsx
```diff
@@ -389,10 +389,13 @@ export function AskUserQuestionCard({
 		if (previousTab.current === tab) return;
 		previousTab.current = tab;
 		const frame = requestAnimationFrame(() => {
-			if (cardRef.current) focusCurrentQuestionPage(cardRef.current);
+			const card = cardRef.current;
+			if (!card) return;
+			actions?.revealChatElement(card, "start");
```
danyaberezun: `tab` and `previousTab` both initialize from the cache, so this reveal never runs after a fresh ChatView remount. Returning to a chat on Q2/review falls back to whole-card `nearest` and can leave the current page start offscreen. Please distinguish a fresh ChatView claim from a same-scope virtualized remount and reveal the restored page.

apps/web/src/chat/tools/AskUserQuestionCard.tsx
```diff
@@ -389,10 +389,13 @@ export function AskUserQuestionCard({
...
+			focusCurrentQuestionPage(card);
```
danyaberezun, line 399: On the review page this focuses Submit/Unanswered after scrolling the card to its start. With a tall recap, `preventScroll` leaves focus offscreen, and the next Enter can submit without the review being visible. Please focus a visible review-start target (or otherwise keep the focus target in view) and add a tall-review case.

apps/web/src/chat/readingBand.ts
```diff
@@ -200,7 +209,34 @@ export function createReadingBandController(
...
+			const measuredBounds = environment.readScrollBounds() ?? bounds;
+			const measuredTarget = latestScrollTop(measuredBounds);
+			const stable =
+				previousTarget !== null &&
+				Math.abs(measuredTarget - previousTarget) <= GEOMETRY_EPSILON &&
+				Math.abs(measuredBounds.scrollTop - measuredTarget) <= GEOMETRY_EPSILON;
+			stableFrames = stable ? stableFrames + 1 : 0;
+			previousTarget = measuredTarget;
+			remainingFrames -= 1;
+			if (stableFrames >= EDGE_PIN_STABLE_FRAMES || remainingFrames <= 0) return;
```
danyaberezun: This stops after three quiet frames (~48 ms), and `setStreaming(false)` also cancels it. A later resize—Mermaid already waits 200 ms—leaves `following` true with Latest hidden even though the viewport is no longer at the edge. Keep the manual-return settle alive for the bounded window and reset it on height changes.

apps/web/src/chat/ChatView.tsx
```diff
@@ -766,15 +784,22 @@ export default function ChatView({
 							key={chatMessageOrder}
 							ref={virtuosoRef}
 							data={rows}
+							heightEstimates={rowHeightEstimates}
 							firstItemIndex={firstItemIndex}
+							increaseViewportBy={CHAT_VIEWPORT_INCREASE}
```
danyaberezun, line 789: `rangeChanged.startIndex` includes the range expanded by `increaseViewportBy`/`minOverscanItemCount` in Virtuoso 4.18.7. `visibleAnchorRowId` can therefore be an offscreen overscan row; a non-prefix newest-first reconciliation then preserves the wrong anchor and jumps. Track the first physically intersecting row and its pixel offset.

danyaberezun (changes requested):
Please address these cases before merging.

danyaberezun (approved):

apps/web/src/chat/rowHeightEstimates.ts
rsolmano: I did not add a general offscreen-size invalidation path in this PR. The Virtuoso lifecycle concern is valid, but we still do not have a deterministic reproduction for the stale persisted-fold/growing-offscreen case, and remounting or rewriting its size tree risks losing the physical reading position. I kept this change scoped to the reproduced failures and would handle invalidation separately once we can pin that case with a regression.

apps/web/src/chat/scrollGeometry.ts
rsolmano: Fixed in 188b1a47. `nearest` is now size-aware: an oversized target below aligns its start, one above aligns its end, and a target spanning both viewport edges does not move. Unit coverage pins both one-sided directions and the spanning case.

apps/web/src/chat/tools/AskUserQuestionCard.tsx
rsolmano: Fixed in 188b1a47. A fresh questionnaire attention claim now repeatedly reveals the restored current page start through the bounded settle window, independently of focus eligibility, so Virtuoso initial placement cannot undo the Q2/review reveal. Pointer, keyboard, or wheel input cancels settling immediately; E2E closes and reopens a persisted Q2 card and verifies its heading and first option are visible.

apps/web/src/chat/tools/AskUserQuestionCard.tsx
rsolmano: Fixed in 188b1a47. Review now lands keyboard focus on the visible review heading rather than an offscreen Submit/Unanswered action. Tab reaches the real footer controls normally, and the tall-review E2E verifies Enter on the heading cannot submit while Submit remains offscreen.

apps/web/src/chat/readingBand.ts
rsolmano: Fixed in 188b1a47 with one bounded variation: manual Latest pinning now stays active for the full 30-frame window and survives `setStreaming(false)`, so delayed growth during that window is corrected. I intentionally did not extend the deadline on each height change, keeping the movement strictly bounded; reader input still cancels immediately. The regression waits ten frames, settles streaming, changes geometry/content, and verifies the new physical edge is pinned before the hard deadline.

apps/web/src/chat/ChatView.tsx
rsolmano: I did not replace the range anchor in this PR. The overscan observation is valid, but we did not obtain a deterministic newest-first non-prefix jump in focused coverage or dogfood, while DOM-intersection plus pixel-offset reconciliation would materially expand this fix. I kept the existing fallback unchanged and would address it separately once there is a pinned reproduction.

PR #366, opened 2026-09-01
fix(terminal): select the Windows shell from ComSpec

lzcer:
## Summary

- preserve an explicit `SHELL` for users who configure one
- select `ComSpec`/`COMSPEC` on Windows, with `cmd.exe` as the final fallback
- keep the existing `/bin/bash` fallback on non-Windows platforms
- keep shell selection local to the terminal module instead of mutating the process environment

## Problem

ThinkRail can be launched from a Windows GUI environment where `SHELL` is not defined. The terminal manager currently falls back to `/bin/bash`, which is not a valid default Windows executable, so `bun-pty` cannot spawn the terminal and the tab reports `could not start a shell`.

## Approach

The terminal module now resolves the PTY executable from the environment and platform. An explicitly configured `SHELL` still wins on every platform. Windows otherwise follows its native `ComSpec` convention, while Unix-like platforms retain the existing bash fallback. This avoids inventing a global `SHELL` value that would also affect the in-process agent and unrelated child processes.

## Validation

- `bun test packages/server/src/terminal/shellArgs.test.ts` (5 pass, 0 fail)
- real Windows `bun-pty` spawn using `C:\Windows\System32\cmd.exe`
- `bun run check:deps`
- `bun run check:boundaries`
- `bun run check:seams`
- `bun run lint` (passes with 6 existing warnings)
- `bun run typecheck`

The complete unit and no-agent E2E suites were also attempted on Windows. Existing fixture/platform assumptions fail before this terminal path is exercised, including `spawnSync mv ENOENT`, missing project fixture rows, JetBrains Central remaining `absent`, temporary-directory `EBUSY`, and POSIX `sleep` usage.


danyaberezun (approved):

danyaberezun (approved):
Approved — ready to merge.

PR #390, opened 2026-09-03
perf(e2e): record local run timings

rsolmano:
## Summary

Add durable, local timing evidence for browser E2E runs so slow, skewed, failed, and interrupted executions can be diagnosed from real history instead of anecdotes. This PR is stacked directly on #389.

## Changes

- Append compact versioned JSONL records to gitignored `e2e/.run-timings.jsonl`, with `THINKRAIL_E2E_TIMING_FILE` as a test/automation override.
- Record mode, raw/effective selection, UTC start, monotonic total/build/shard/phase durations, outcome, exit code, and full-run parent lineage across source, agent, full, binary, and desktop modes.
- Start timing in the shared preload so idle-sleep setup is included and its failures are captured; suppress true list/preflight commands with option-aware parsing.
- Keep writes best-effort and ensure managed interruptions include bounded descendant cleanup before finalization.
- Harden full/Central list preflights and clear desktop timeout/readiness losers so diagnostics cannot alter or underreport the runner lifecycle.
- Add focused timing, lineage, interruption, and environment tests plus spec/docs updates.

## Stack

1. #389 — macOS idle-sleep guard
2. **local JSONL timing records (this PR)**
3. #391 — focused and repair commands

Land bottom-up in this order.

## Testing

- `bun run e2e` — 350 passed across 8 shards
- `bun run check:deps && bun run check:boundaries && bun run check:seams && bun run lint && bun run typecheck && bun run test` — passed
- Focused timing/agent/signal coverage — 18 passed
- Focused `e2e:full` lineage run — 6 passed with linked parent/child records
- True `--list` runs produced no timing file; `--grep --list` remained a recorded non-list failure

PR #391, opened 2026-09-03
perf(e2e): add focused repair commands

rsolmano:
## Summary

Make the cheap repair loop explicit without weakening or changing the complete `bun run e2e` handoff gate. This PR is stacked directly on #390.

## Changes

- Add `e2e:focused` for one-host, one-worker iteration that stops after the first failure.
- Add `e2e:repair` for a serial `--last-failed` run that deliberately omits fail-fast, ensuring every remembered failure is revisited before Playwright replaces its last-run file.
- Keep the existing `e2e` script byte-for-byte unchanged as the complete machine-adaptive no-agent gate.
- Pin the command semantics in runner coverage and update the E2E contract and contributor guidance.

## Stack

1. #389 — macOS idle-sleep guard
2. #390 — local JSONL timing records
3. **focused and repair commands (this PR)**

Land bottom-up in this order.

## Testing

- `bun run e2e` — 351 passed across 8 shards
- `bun run check:deps && bun run check:boundaries && bun run check:seams && bun run lint && bun run typecheck && bun run test` — passed
- `bun run e2e:focused -- e2e/shard-runner.spec.ts` — 8 passed; timing confirmed one shard plus `--max-failures=1`
- Synthetic two-failure run followed by `bun run e2e:repair` — both remembered failures executed
- Post-rebase focused cumulative runner coverage — 26 passed

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -155,8 +155,8 @@ reusable by any pi UI (extraction-ready as a future `packages/chat-ui`).
 
 Every change that touches the app is verified by the **complete e2e suite once before it is considered
 done**. During implementation, iterate with the affected spec
-(`bun run e2e -- e2e/<feature>.spec.ts`) or `bun run e2e -- --last-failed`; do not rerun the full gate
-after every edit.
+(`bun run e2e:focused -- e2e/<feature>.spec.ts`) or `bun run e2e:repair`; do not rerun the full gate after
+every edit. Focused runs stop after the first failure; repair runs every failure remembered by Playwright.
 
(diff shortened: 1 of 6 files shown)


rsolmano:
Closing this after a fresh end-to-end review of the stack.

Focused spec execution and `--last-failed` already existed, and the repository already instructed agents to use both. This PR mostly adds two aliases; the only material behavior change is adding `--max-failures=1` to one of them. That convenience does not justify a standalone performance PR or the surrounding stack complexity.

The retained PR is #389, which addresses the measured macOS sleep problem directly. If agent iteration behavior remains a demonstrated problem, it can be revisited separately as a small workflow change with evidence of the effect.

PR #397, opened 2026-09-03
docs(contracts): restructure the wire spec for the readability gate (pilot)

rsolmano:
## Summary

**Stacked on #387** (the warn-only `check:specs` gate) — review/merge that first.

Pilot rewrite of the spec-readability program: `packages/contracts/SPEC.md` (the corpus' most decision-dense spec, 5,628 words) restructured to the new skeleton and gate budgets. Every decision, invariant, rejected alternative, version note, and PR/issue anchor is preserved — this was a compression + restructuring pass, never delete-to-fit.

**5,628 → 4,489 words** · 20 section headings · worst prose block 77 words (was 4,977) · zero `check:specs` violations.

## What changed

- **Skeleton by decision area** instead of a per-file transcript: error codes · pi mirrors · WireModel · session read shapes · ask/subagent wire · domain entities · auth+Central wire · skills+trust · TODO plan · review · specs+history · templates · layout+config · WS methods · WS channels · request idempotency · Get right · Consumed by.
- **`spec-budget: 4500` override** (visible frontmatter, hard ceiling per the gate) with in-body rationale: the wire is *one genuine boundary* — splitting it to fit 3,000 would scatter one boundary's truth across files. This is the sanctioned escape valve, and the pilot's evidence that it belongs on single-boundary decision-dense specs only; collections (panels, chat) will split instead.
- **`scripts/SPEC.md`**: the checker paragraph split to satisfy the block cap.

## Preservation protocol (pilot calibration)

1. Worker subagent: decision-granularity inventory → rewrite → self-check. (Calibration finding: inventory must be written to a file, not returned in chat — the worker's report truncated. Amended for the giants.)
2. **Author mechanical aid**: backtick-token diff old-vs-new — caught one real loss (`ask_user_question`, the tool name; restored) and confirmed every other token delta was a deliberate qualified form (`Workspace.renamed`, `review.get`, …).
3. **Author section-by-section pass**: complete; no gaps found (all wire members, mirrors, v43/v55/v56 notes, #97/#303 anchors, bundle gate, extensionless-imports note, `Model<any>` present).
4. **Independent checker subagent** (original-vs-new, no worker mapping): verdict pending — will be posted here when it lands; this PR stays draft until then.

## Testing

- `bun scripts/check-specs.ts --verbose` — contracts: **4489/4500 words, 20 headings, worst block 77, no violations**
- `bun run check:spec-surface` — 14 enrolled / 14 compared (enrollment unchanged)
- `bun test packages/contracts scripts/` — 71 pass
- Pre-commit fast gates green (deps, boundaries, seams, lint, typecheck)
- Docs-only change: no product-runtime path; local e2e intentionally not run


rsolmano:
**Independent checker verdict (original-vs-new, no worker mapping): 119/120 inventory items fully preserved; 0 missing; 0 NEW-only drift.**

The one finding — a **weakened** clause, not a loss: the user-owned-workspace `baseBranch` rule survived, but its *reason* was dropped ("whose provenance isn't ThinkRail's to claim"). Without it a future implementer could "fix" the special case into misattributing provenance. Restored in 89d6cbd6.

Every named anchor verified present: v43/v55/v56, 0.79.8–0.79.9, #97, #303, architecture #9; all mirror-vs-import rules + rationales; the three shared-reading single-source rules with jump-anchor alignment; WireModel fails-closed + host re-resolution; `agent_settled` vs `agent_end`; both queue shapes; ack+terminate; full method/channel inventories with per-member edge semantics; WsAck/WsResume + why-resume.

Protocol note: this is exactly the catch class the three layers were designed for — worker self-check missed it (prose rationale, not a token), the mechanical token-diff couldn't see it, the independent checker caught it.

PR #402, opened 2026-09-03
tests: fixture repos do not inherit the developer's signing config

CommanderTvis:
# Problem

Two git tests set `commit.gpgsign` with `gpg.program` pointed at a binary that does not exist and expect the commit to fail. On a machine that signs with `gpg.format=ssh` (my case: 1Password signing) the bogus program is never consulted: git reaches for the ssh signer instead, the commit succeeds, and the tests fail.

That uncovered the larger half. Every fixture repo inherits the developer's global git config, so with `commit.gpgsign=true` globally the whole suite was signing each seed commit for real. That is an agent round-trip per fixture, and with a hardware-backed key it is a prompt in the middle of a test run. The e2e spec helpers and the artifact smoke probe run with the developer's real HOME (only the host under test gets the isolated one), so they inherit the same way the unit-test fixtures do.

# Change

- The two failing-commit tests pin `gpg.format=openpgp` alongside `gpg.program`, so the program they broke is the one git uses.
- The `initProject` tests cannot pre-seed repo config, since `initProject` runs `git init` and `git commit` itself. Every call now goes through a `withoutUserGitConfig` helper that points `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` at `/dev/null`, the isolation the no-identity test already used.
- Every fixture seed sets `commit.gpgsign=false`: the unit-test repos under `packages/server`, the e2e fixture repo and the workflow-harness workspace, the second-project seed in the projects e2e spec, the ask-restart live seed, and the artifact smoke probe's fixture project. Nothing in these repos is ever pushed, and a test repository in a temp directory has nothing to attest.

# Verification

- `bun run check:deps`, `check:boundaries`, `check:seams`, `lint`, `typecheck`: green.
- `packages/server` unit suite: 925 pass, 0 fail, on a machine with `commit.gpgsign=true` and `gpg.format=ssh`.
- `packages/server` unit suite with `GIT_CONFIG_GLOBAL` pointed at an unsignable config (`commit.gpgsign=true`, `gpg.program=/nonexistent`): 925 pass, 0 fail.
- `bun run e2e` (full no-agent suite): 344 passed. A first run had one failure in the `projects` spec at the `welcome-title` assertion; it passed on rerun, both alone and in a full clean run, and the seed it follows had already completed.

### DIFF
--- a/e2e/ask-restart.live.spec.ts
+++ b/e2e/ask-restart.live.spec.ts
@@ -63,6 +63,7 @@ function seedState(): void {
 	gitQuiet(REPO, "init", "-b", "main");
 	gitQuiet(REPO, "config", "user.email", "e2e@thinkrail.test");
 	gitQuiet(REPO, "config", "user.name", "ThinkRail E2E");
+	gitQuiet(REPO, "config", "commit.gpgsign", "false");
 	writeFileSync(join(REPO, "README.md"), "# restart fixture\n");
 	gitQuiet(REPO, "add", "-A");
 	gitQuiet(REPO, "commit", "-m", "init");
--- a/e2e/fixtures/repo.ts
+++ b/e2e/fixtures/repo.ts
@@ -18,6 +18,7 @@ export function seedFixtureRepo(): void {
 	git("init", "-b", "main");
 	git("config", "user.email", "e2e@thinkrail.test");
 	git("config", "user.name", "ThinkRail E2E");
+	git("config", "commit.gpgsign", "false");
 	writeFileSync(join(E2E_FIXTURE_REPO, "README.md"), "# sample-project\n");
 	writeFileSync(join(E2E_FIXTURE_REPO, "notes.txt"), "plain-text-fixture\n");
 	writeFileSync(
--- a/e2e/projects.spec.ts
+++ b/e2e/projects.spec.ts
@@ -103,6 +103,7 @@ function seedSecondRepo(): string {
 	git(repo, "init", "-b", "main");
 	git(repo, "config", "user.email", "e2e@thinkrail.test");
 	git(repo, "config", "user.name", "ThinkRail E2E");
+	git(repo, "config", "commit.gpgsign", "false");
 	writeFileSync(join(repo, "README.md"), "# second project\n");
 	git(repo, "add", "-A");
 	git(repo, "commit", "-m", "seed");
--- a/e2e/workflows/harness/workspace.ts
+++ b/e2e/workflows/harness/workspace.ts
@@ -14,6 +14,7 @@ export function seedWorkspace(seed: WorkspaceSeed): string {
 	gitQuiet(cwd, "init", "-b", "main");
 	gitQuiet(cwd, "config", "user.email", "workflow-tests@thinkrail.test");
 	gitQuiet(cwd, "config", "user.name", "ThinkRail Workflow Tests");
+	gitQuiet(cwd, "config", "commit.gpgsign", "false");
 	if (typeof seed === "function") seed(cwd);
 	else seedKind(cwd, seed);
 	gitQuiet(cwd, "add", "-A");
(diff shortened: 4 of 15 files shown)


danyaberezun (changes requested):
Two tests are missed and still sign with the developer's key: `initProject: initialises a plain folder…` and `initProject: an empty folder gets an empty initial commit…` in `packages/server/src/projects/projects.test.ts`. `initProject` runs `git init` + `git commit` itself, so repo-local config can't be pre-seeded. Verified by running the server suite with `GIT_CONFIG_GLOBAL` pointing at an unsignable config (`commit.gpgsign=true`, `gpg.program=/nonexistent`): these are the only two failures on the branch.

Reuse the isolation the `-c fallback` test already does:

```diff
+function withoutUserGitConfig<T>(fn: () => T): T {
+	const savedGlobal = process.env.GIT_CONFIG_GLOBAL;
+	const savedSystem = process.env.GIT_CONFIG_SYSTEM;
+	process.env.GIT_CONFIG_GLOBAL = "/dev/null";
+	process.env.GIT_CONFIG_SYSTEM = "/dev/null";
+	try {
+		return fn();
+	} finally {
+		if (savedGlobal === undefined) delete process.env.GIT_CONFIG_GLOBAL;
+		else process.env.GIT_CONFIG_GLOBAL = savedGlobal;
+		if (savedSystem === undefined) delete process.env.GIT_CONFIG_SYSTEM;
+		else process.env.GIT_CONFIG_SYSTEM = savedSystem;
+	}
+}

 test("initProject: initialises a plain folder, commits its contents, and opens it", () => {
 	…
-	const project = initProject(dir);
+	const project = withoutUserGitConfig(() => initProject(dir));

 test("initProject: an empty folder gets an empty initial commit (a HEAD), so worktrees work", () => {
 	…
-	initProject(dir);
+	withoutUserGitConfig(() => initProject(dir));

 test("initProject: commits even with no configured git identity (the -c fallback)", () => {
 	…
-	const savedGlobal = process.env.GIT_CONFIG_GLOBAL;
-	const savedSystem = process.env.GIT_CONFIG_SYSTEM;
-	process.env.GIT_CONFIG_GLOBAL = "/dev/null";
-	process.env.GIT_CONFIG_SYSTEM = "/dev/null";
-	try {
-		initProject(dir);
-		expect(gitOut(dir, "rev-parse", "HEAD")).not.toBe("");
-		expect(gitOut(dir, "log", "-1", "--format=%an")).toBe("ThinkRail");
-	} finally {
-		if (savedGlobal === undefined) delete process.env.GIT_CONFIG_GLOBAL;
-		else process.env.GIT_CONFIG_GLOBAL = savedGlobal;
-		if (savedSystem === undefined) delete process.env.GIT_CONFIG_SYSTEM;
-		else process.env.GIT_CONFIG_SYSTEM = savedSystem;
-	}
+	withoutUserGitConfig(() => initProject(dir));
+	expect(gitOut(dir, "rev-parse", "HEAD")).not.toBe("");
+	expect(gitOut(dir, "log", "-1", "--format=%an")).toBe("ThinkRail");
```

CommanderTvis:
> The neighbouring test ("commits even with no configured git identity") already does the right thing with GIT_CONFIG_GLOBAL=/dev/null + GIT_CONFIG_SYSTEM=/dev/null. Could you pull that save/set/restore into a small helper and wrap all three?

Thanks for pointing it out. Sorry for not paying attention to the neighboring tests at all. For me, this PR is just an unblocker for the constant env-dependent e2e failures.

CommanderTvis:
Claude:

Pushed 77b1342d with `withoutUserGitConfig` wrapping the three tests.

Repeating your `GIT_CONFIG_GLOBAL` check on the branch turned up one more call: the host-home test also runs `initProject`. It swaps `HOME`, so it is isolated whenever the config lives in `~/.gitconfig`, but not when it comes in through `GIT_CONFIG_GLOBAL`, which is why it passed for you. That call goes through the helper as well.

Server suite: 925/925 under the unsignable global config and under my real one (`gpg.format=ssh`, `commit.gpgsign=true`).

danyaberezun (approved):

PR #409, opened 2026-09-04
fix(git): branch pickers enumerate every remote, not only origin

### LINKED ISSUE #371: Branch picker has to enumerate all remotes
## Problem

When choosing branch in 'Compare' pane, I see only `origin` remote.

When developing a fork, you almost always work with two remotes.

## Proposed solution

All remotes displayed in all branch enumerations.

## Additional context

CommanderTvis:
Closes #371.

# Problem

The branch pickers listed `refs/remotes/origin` and nothing else. On a fork, where the interesting base is `upstream/main`, that ref was not offered at all. The hard-coded `origin/` reached past the list too: prefetch and workspace-create both fetched from `origin` whatever ref they were handed, and remote-ness was decided by the string's prefix, which cannot tell `upstream/main` from a local branch called `feature/main`.

# Change

- `listBranches` enumerates all direct refs under `refs/remotes`, omits symbolic aliases, and keeps namespace-stripped full refs such as `origin/main` and `upstream/main` as canonical identity. Optional host-authored grouping metadata lets the picker render `Remote → origin/upstream → branch` without parsing refs in the browser; older hosts retain the flat full-ref fallback, and legal remote names containing `/` stay unambiguous.
- Remote-ness is decided against the repository's configured remotes. `remoteNameOf(ref, remotes)` chooses the longest matching prefix, and prefetch/workspace-create fetch from that remote and check out the fully-qualified tracking ref.
- Initial default prefetch is always offered to the host, which safely no-ops local refs. This restores background refresh for both stale and missing default tracking refs; manual picks prefetch only rows from the loaded remote list.
- `resolveDefaultBranch` keeps its sync spawn count while reading full ref names: origin's dangling `HEAD` still works, only `/HEAD` symrefs can provide another remote default, and a local branch named `origin/main` cannot hide the exact remote ref.
- `remoteRefOid` asks only whether the full tracking ref resolves, including one that outlived its remote configuration.
- `pr.open` rejects non-origin remote bases before reading dirty files or pushing. Cross-repository fork PR creation remains explicit future scope rather than publishing a branch before discovering an invalid target.
- `countUnpushedCommits` deliberately keeps its `origin/`: it asks where this workspace's own branch was pushed, not which remote its base came from.

# Provenance

Cherry-picked from `469d2f1` on my `claude-code-integration` branch, then reviewed and reworked on top of upstream main. The spawn-neutral default resolver remains: adding a sync process to every Default-workspace ensure widened a pre-existing e2e state-reset race, while one full-ref `for-each-ref` preserves the old process budget.

Review hardening landed in `c7d9c41a`: it removed browser-side remote-name parsing, repaired default prefetch and full-ref resolution, added the pre-push non-origin guard, aligned the owning specs, and added real Git plus browser regressions for the discovered cases. `4c125e17` adds the reviewer-requested two-layer presentation through additive host-owned metadata while preserving full refs for selection, search, and wire identity.

# Verification

- `check:deps`, `check:boundaries`, `check:seams`, `check:spec-surface`, `lint`, `typecheck`: green.
- `bun run test`: 14/14 workspaces green.
- `bun run e2e`: 351 tests passed across all 8 shards.

## Screenshots

| Before — only `origin` | After — grouped remotes |
| --- | --- |
| Before: branch picker only lists origin | After: branch picker groups origin and upstream into separate remote sections |

https://claude.ai/code/session_01DZ4jbu6P52ZGptoEgzqsYf


CommanderTvis:
@danyaberezun I think that it's better to have two-layer grouping

image

has more clutter (due to repetitive remote name, `origin/`) than, for example, IDEA's

image

CommanderTvis:
Converting to draft until we have decided between two-layer (remote or local -> remote -> branch) grouping vs one-layer (remote or local -> `remote/branch`) grouping.

danyaberezun:
Implemented the two-layer grouping in [`4c125e17`](https://github.com/JetBrains/thinkrail/commit/4c125e17e960952b806045ab76e200a384ea1387).

The host now supplies additive `remoteGroups` metadata, and the picker renders **Remote → remote → branch**. Full refs remain the selection/search/wire identity, so the browser never splits refs (including remotes such as `team/upstream`), while newer clients still fall back to the flat full-ref list against older hosts. I also refreshed the PR screenshot with the final implementation.

The grouping decision is resolved, so I’m marking the PR ready for review again.

CommanderTvis:
LGTM!

danyaberezun (approved):

CommanderTvis:
I pushed CI fixup

rsolmano (approved):

PR #413, opened 2026-09-04
feat(web): show agent activity in the Projects rail

### LINKED ISSUE #273: Show per-workspace agent status: working / needs you / error / idle
## Problem

With several workspaces open, the Projects Rail tells you nothing about what's happening in the ones
you aren't looking at. Every row looks the same whether the agent is mid-turn, blocked waiting on you,
errored, or idle.

From the team discussion:

> There's a big problem — you have no idea what's going on in which workspace. Where the agent is
> working, where it's waiting on some action from the user, and where it's just sitting idle.
>
> At minimum just an indicator in the menu that something is needed from the user, or that there was
> an error.

The costly case is **blocked**: a run stopped on an `ask_user_question` card or an `extensionUi`
dialog sits there indefinitely, because nothing outside that chat tab says it's your turn. Errors are
equally invisible from outside the tab. Parallel-agent work across worktrees is the point of the
product, and today it only scales as far as you can keep track of in your head.

## Proposed solution

Show a small per-workspace status indicator on each workspace row in the Projects Rail, rolled up onto
the project row when collapsed — for every workspace, not just the active one.

Extend the existing glance vocabulary (`PlanGlance` in `apps/web/src/chat/planView.ts`, icons in
`chat/TodoList.tsx`) rather than inventing a parallel one:

| State | Source |
|---|---|
| **Needs you** | An unanswered `ask_user_question`, or a pending `extensionUi` dialog. |
| **Error** | New state — `AgentSettlement.stopReason` is an error. |
| **Working** | Streaming, or queued work not yet drained. |
| **Idle** | Has chats, none of the above. |

- Priority when a workspace has several chats: **Needs you > Error > Working > Idle**.
- "Working" must mean *not yet settled*, not `agent_end` (see #131).
- Glyph + semantic token, not colour alone.

Out of scope here: OS/push notifications, sound, a global cross-project inbox, cost badges.

## What exists today

- `SessionSummary` already carries `workspaceId`, `isStreaming`, `live`, `lastSettlement`
  (`packages/contracts/src/piProtocol.ts:101`). It's a `session.list` response, never pushed.
- `PlanGlance` / `sessionGlance()` already derive `working | waiting_question | waiting` per session;
  `glanceIcon` already has icons and labels. **No `error` state exists.**
- `pi.event` is broadcast globally, so a *visited* workspace's sessions keep updating in the store
  even in the background.

The gaps:

1. **`isStreaming` is not enough.** The ask tool acks and ends the turn, so `isStreaming` is `false`
   while a question is pending. "Needs you" has to be derived from the transcript
   (`chat/askState.ts`) or promoted to a real field on the wire.
2. **Unvisited workspaces are invisible.** `session.list` runs only for the active workspace, and
   `SessionEventPayload` has no `workspaceId`. A rail indicator needs a host-side per-workspace status
   push (or an eager status read) — not eager hydration of every transcript.
3. Reaching `glanceIcon` from `panels/` would create a `panels → chat` import edge; the map likely
   moves to a neutral module.

## Alternatives considered

- **Status on chat tabs only** — doesn't help; you must already be in the workspace to see it.
- **Toasts on transitions** — missable, and can't answer "what's the state of everything right now".
- **Eagerly hydrate every workspace's sessions** — pays full transcript cost to render a dot.

## Related

- #131 — `agent_settled`; the correctness constraint for "Working".
- #241 — in-session error handling; overlaps with what "Error" means here.
- #240 removed `+N −M` badges from workspace rows on the grounds that "the Projects view is for
  navigation and identity" (`apps/web/src/panels/SPEC.md`). This argues against that rule and should
  amend it: attention *is* navigation. Keep the rail restrained — one indicator, not a badge cluster.
- Nearest precedent for a selector-fed sidebar badge: the skills-reload badge
  (`selectSkillsStale`, `apps/web/src/store/SPEC.md`).
- Specs to amend: `apps/web/src/panels/SPEC.md`, `apps/web/src/store/SPEC.md`,
  `packages/contracts/SPEC.md`, `packages/server/src/agent/SPEC.md`, `apps/web/UI-TERMINOLOGY.md`.
- Design input wanted on the visual treatment before implementation.

rsolmano:
## Summary

With several workspaces each running their own agents, the Projects rail was the only always-visible
place that listed them — and it said nothing about what was happening inside. The only way to find out
was to open every workspace in turn.

Each workspace row now carries a glyph for what its agents are doing, rolled up across its chats, with
an explanation on hover. Collapsed project rows carry the rollup too. Idle draws **nothing**, so a quiet
rail is visually identical to before.

Five states, derived host-side: `running`, `waiting` (a question or a blocking dialog is on you),
`queued`, `failed`, and idle. A run **you** cancelled is idle, not failed — a red row for every Escape
would teach you to ignore the signal.

Two dimensions were deliberately collapsed into one. `failed` and `waiting` are *durable derived states*,
not moments: a failure stays in the settlement until the next run, and an unanswered question stays
unanswered until answered. So the states that need to persist persist by construction — no seen/unseen
layer, no acknowledgement, no clearing logic.

Notifications (toasts, native) are **out of scope** and deliberately deferred.

## Screenshots

The same workspace with a failed chat, before and after:

| Before | After |
|---|---|
| before | after |

An unanswered question, where *you* are the blocker:

waiting

## Changes

**`packages/contracts`** — `ActivityStatus` + `SessionActivity`, the `session.activity` push channel, the
`session.activityList` snapshot read, and `ACTIVITY_PROTOCOL_VERSION` (protocol → 59). `idle` is **not a
union member**: it is absence everywhere, so quiet is the default a consumer gets for free rather than a
value each one must remember to special-case. Also **deletes** `TabStatus` and `Session` — unreferenced
anywhere in the repo, a vestigial first attempt at exactly this feature.

**`packages/server`** — `agent/activity.ts` derives one session's status as a pure function; the manager
owns only publish-on-change, the snapshot, and retraction on disposal. Order is dialog > running > queued
> waiting > failed, and every rung is load-bearing (see `agent/SPEC.md`): a blocking dialog outranks
streaming because pi is mid-turn while *you* are the blocker, and `queued` outranks `waiting` because a
queued message is not yet in the transcript, so answerability cannot see it. `waiting` and `failed` both
fall back to the transcript, so they survive a host restart for any attached session.

**`apps/web`** — `activityByWorkspace` (workspace → session → status), fed by the push with a snapshot
read on connect, gated on the protocol version. The rollup is derived on read by pure
`workspaceActivityRollup`/`projectActivityRollup` rather than Zustand selectors: a fresh rollup object
returned from a selector would re-render the whole rail on every store change. `ActivityGlyph` renders
Remix line icons on semantic tokens.

**Not per-chat.** The status is per session on the wire but the rail shows one glyph per row: how several
chats collapse into one row is presentation policy, and a precedence order baked into the host would be a
UI decision escaping into the engine. Notably the rollup order (`failed` first) is deliberately *not* the
host's per-session order — different questions, and both are pinned by tests so neither gets "fixed" into
the other.

**Design notes.** Icons rather than coloured dots because five states encoded purely in hue fail
colour-blind users and the shipped high-contrast themes. `running` is `feedback-info`, never the accent —
the active workspace's row already renders in the accent, so a green glyph would read as selection. No
motion: the rail is permanently in peripheral vision and concurrent runs pulsing out of phase read as
flicker (the chat plan pane keeps its pulse, since that surface is actively read).

`apps/web/src/panels/SPEC.md` recorded that workspace rows carry no badges. That decision is **amended
explicitly**, not quietly contradicted: the rail admits exactly one decoration class — live agent state,
because it answers a question the rail is the only place to ask — and the rule still excludes change
detail such as a `+N −M` badge.

## Testing

- `bun run e2e` — all 8 shards passed in 145.1s (includes 4 new specs in `e2e/workspace-activity.spec.ts`)
- `bun run test` — 14/14 workspaces, 0 fail (server 950, web 918; 39 new tests across the activity
  derivation, the store fold, the rollup, and the glyph)
- `bun run typecheck` — 14/14 green
- `bun run lint` — clean (6 pre-existing warnings, unchanged)
- `bun run check:deps` / `check:boundaries` / `check:seams` / `check:spec-surface` — all OK

The E2E runs on the **no-agent** lane with no provider spend: a seeded fixture transcript becomes real
activity the moment opening the chat attaches the session, so the full chain — host derivation, push,
store fold, rollup, render — is exercised for real. That is also why `failed` falls back to the
transcript; writing the test is what surfaced the asymmetry with `waiting`.

### DIFF
--- a/apps/web/src/panels/ActivityGlyph.tsx
+++ b/apps/web/src/panels/ActivityGlyph.tsx
@@ -39,6 +39,10 @@ const PRESENTATION: Record<
 
 const BREAKDOWN_ORDER: readonly ActivityStatus[] = ["failed", "waiting", "running", "queued"];
 
+export function activityChatCount(counts: Partial<Record<ActivityStatus, number>>): number {
+	return BREAKDOWN_ORDER.reduce((total, status) => total + (counts[status] ?? 0), 0);
+}
+
 export function activityBreakdown(counts: Partial<Record<ActivityStatus, number>>): string[] {
 	return BREAKDOWN_ORDER.flatMap((status) => {
 		const count = counts[status] ?? 0;
@@ -56,7 +60,7 @@ export function ActivityGlyph({
 }) {
 	const { Icon, label, className } = PRESENTATION[status];
 	const lines = counts ? activityBreakdown(counts) : [];
-	const detailed = lines.length > 1;
+	const detailed = counts !== undefined && activityChatCount(counts) > 1;
 	return (
 		<IconTooltip
 			wrapTrigger
--- a/apps/web/src/panels/ActivityGlyph.tsx
+++ b/apps/web/src/panels/ActivityGlyph.tsx
@@ -1,0 +1,85 @@
+import {
+	RiErrorWarningLine as CircleAlert,
+	RiRecordCircleLine as CircleDot,
+	RiQuestionnaireLine as MessageCircleQuestion,
+	RiTimeLine as Timer,
+} from "@remixicon/react";
+import type { ActivityStatus } from "@thinkrail/contracts";
+import { IconTooltip } from "@/components/ui/tooltip";
+
+const PRESENTATION: Record<
+	ActivityStatus,
+	{ Icon: typeof CircleDot; label: string; plural: string; className: string }
+> = {
+	failed: {
+		Icon: CircleAlert,
+		label: "Last run failed",
+		plural: "failed",
+		className: "text-feedback-error",
+	},
+	waiting: {
+		Icon: MessageCircleQuestion,
+		label: "Waiting for your answer",
+		plural: "waiting for your answer",
+		className: "text-feedback-warning",
+	},
+	running: {
+		Icon: CircleDot,
+		label: "Agent is working",
+		plural: "working",
+		className: "text-feedback-info",
+	},
+	queued: {
+		Icon: Timer,
+		label: "Message queued",
+		plural: "queued",
+		className: "text-text-subtle",
+	},
+};
+
+const BREAKDOWN_ORDER: readonly ActivityStatus[] = ["failed", "waiting", "running", "queued"];
+
+export function activityBreakdown(counts: Partial<Record<ActivityStatus, number>>): string[] {
+	return BREAKDOWN_ORDER.flatMap((status) => {
+		const count = counts[status] ?? 0;
+		if (count === 0) return [];
+		return [`${count} ${count === 1 ? "chat" : "chats"} ${PRESENTATION[status].plural}`];
+	});
+}
+
+export function ActivityGlyph({
+	status,
+	counts,
+}: {
+	status: ActivityStatus;
+	counts?: Partial<Record<ActivityStatus, number>>;
+}) {
+	const { Icon, label, className } = PRESENTATION[status];
+	const lines = counts ? activityBreakdown(counts) : [];
+	const detailed = lines.length > 1;
+	return (
+		<IconTooltip
+			wrapTrigger
+			label={
+				detailed ? (
+					<span className="flex flex-col gap-2">
+						{lines.map((line) => (
+							<span key={line}>{line}</span>
+						))}
+					</span>
+				) : (
+					label
+				)
+			}
+		>
+			<span
+				data-testid="activity-glyph"
+				role="img"
+				aria-label={detailed ? lines.join(", ") : label}
+				className="flex size-20 shrink-0 items-center justify-center"
+			>
+				<Icon className={`size-14 ${className}`} />
+			</span>
+		</IconTooltip>
+	);
+}
--- a/apps/web/src/panels/ProjectTree.tsx
+++ b/apps/web/src/panels/ProjectTree.tsx
@@ -397,6 +397,7 @@ function ProjectRow({
 		<div
 			data-testid="project-item"
 			data-menu-open={menuOpen}
+			{...(activity ? { "data-activity": activity.status } : {})}
 			className={`group flex h-28 items-center gap-4 rounded-[var(--radius-sm)] pr-4 pl-4 transition-colors ${
 				menuOpen ? "bg-control-bg-selected" : "hover:bg-control-bg-hovered"
 			}`}
--- a/apps/web/src/panels/ProjectTree.tsx
+++ b/apps/web/src/panels/ProjectTree.tsx
@@ -48,13 +48,17 @@ import {
 import { copyText } from "@/lib";
 import { LoadingRegion } from "../components/Skeleton";
 import {
+	type ActivityRollup,
 	isDefaultWorkspace,
 	isExternalWorkspace,
+	projectActivityRollup,
 	selectActiveWorkspaceProjectId,
 	toast,
 	useAppStore,
+	workspaceActivityRollup,
 } from "../store";
 import { errorText, getTransport, prewarmWorkspaceSkillLoad } from "../transport";
+import { ActivityGlyph } from "./ActivityGlyph";
 import { AddProjectMenu } from "./AddProjectMenu";
 import { ConfirmDialog } from "./ConfirmDialog";
 import { ExistingWorktreeDialog } from "./ExistingWorktreeDialog";
@@ -71,6 +75,7 @@ export function ProjectTree() {
 	const workspaces = useAppStore((s) => s.workspaces);
 	const worktreeCreations = useAppStore((s) => s.worktreeCreationsByProject);
 	const activeWorkspaceId = useAppStore((s) => s.activeWorkspaceId);
+	const activityByWorkspace = useAppStore((s) => s.activityByWorkspace);
 	const protocolVersion = useAppStore((s) => s.protocolVersion);
 
 	const [editors, setEditors] = useState<EditorInfo[]>([]);
@@ -263,6 +268,11 @@ export function ProjectTree() {
 								project={project}
 								isSelected={selectedProjectId === project.id}
 								isExpanded={isExpanded}
+								activity={
+									isExpanded
+										? null
+										: projectActivityRollup(activityByWorkspace, workspaces, project.id)
+								}
 								workspaceCount={(list ?? []).filter((w) => !isDefaultWorkspace(w)).length}
 								onToggle={() => toggleExpand(project.id)}
 								onSelect={() => void selectProject(project.id)}
@@ -283,6 +293,7 @@ export function ProjectTree() {
 											key={ws.id}
 											workspace={ws}
 											isActive={activeWorkspaceId === ws.id}
+											activity={workspaceActivityRollup(activityByWorkspace, ws.id)}
 											canRename={canRenameWorkspace(protocolVersion, ws)}
 											editors={editors}
 											onSelect={() => selectWorkspace(ws)}
@@ -348,6 +359,7 @@ function ProjectRow({
 	project,
 	isSelected,
 	isExpanded,
+	activity,
 	workspaceCount,
 	onToggle,
 	onSelect,
@@ -361,6 +373,7 @@ function ProjectRow({
 	project: Project;
 	isSelected: boolean;
 	isExpanded: boolean;
+	activity: ActivityRollup | null;
 	workspaceCount: number;
 	onToggle: () => void;
 	onSelect: () => void;
@@ -414,6 +427,7 @@ function ProjectRow({
 					{project.name}
 				</span>
 			</button>
+			{activity && <ActivityGlyph status={activity.status} counts={activity.counts} />}
 			{!isExpanded && workspaceCount > 0 && (
 				<span
 					data-testid="project-workspace-count"
@@ -518,6 +532,7 @@ function ProjectRow({
 function WorkspaceRow({
 	workspace,
 	isActive,
+	activity,
 	canRename,
 	editors,
 	onSelect,
@@ -529,6 +544,7 @@ function WorkspaceRow({
 }: {
 	workspace: Workspace;
 	isActive: boolean;
+	activity: ActivityRollup | null;
 	canRename: boolean;
 	editors: EditorInfo[];
 	onSelect: () => void;
@@ -638,6 +654,7 @@ function WorkspaceRow({
 				data-testid="workspace-item"
 				data-active={isActive}
 				data-kind={workspace.kind ?? "worktree"}
+				{...(activity ? { "data-activity": activity.status } : {})}
 				onContextMenu={openMenuFromContext}
 				className={`group flex min-h-28 min-w-0 items-center gap-8 rounded-[var(--radius-sm)] border-0 py-4 pr-4 pl-24 transition-colors ${
 					isActive || menuOpen ? "bg-control-bg-selected" : "hover:bg-control-bg-hovered"
@@ -676,6 +693,7 @@ function WorkspaceRow({
 						</span>
 					</button>
 				)}
+				{activity && <ActivityGlyph status={activity.status} counts={activity.counts} />}
 				<DropdownMenu open={menuOpen} onOpenChange={setMenuOpen}>
 					<DropdownMenuTrigger
 						data-testid="workspace-menu"
@@ -754,6 +772,7 @@ function WorkspaceRow({
 					</DropdownMenuContent>
 				</DropdownMenu>
 			</fieldset>
+
 			{!isDefault && (
 				<ConfirmDialog
 					open={confirmOpen}
--- a/apps/web/src/panels/ProjectTree.tsx
+++ b/apps/web/src/panels/ProjectTree.tsx
@@ -269,9 +269,7 @@ export function ProjectTree() {
 								isSelected={selectedProjectId === project.id}
 								isExpanded={isExpanded}
 								activity={
-									isExpanded
-										? null
-										: projectActivityRollup(activityByWorkspace, workspaces, project.id)
+									isExpanded ? null : projectActivityRollup(activityByWorkspace, project.id)
 								}
 								workspaceCount={(list ?? []).filter((w) => !isDefaultWorkspace(w)).length}
 								onToggle={() => toggleExpand(project.id)}
--- a/apps/web/src/panels/ProjectTree.tsx
+++ b/apps/web/src/panels/ProjectTree.tsx
@@ -237,7 +237,7 @@ export function ProjectTree() {
 	};
 
 	return (
-		<nav className="flex flex-col gap-8">
+		<nav data-testid="project-tree" className="flex flex-col gap-8">
 			<header className="flex h-28 items-center justify-between pr-4 pl-8">
 				<span className="tr-text-eyebrow text-text-muted">Projects</span>
 				<AddProjectMenu
--- a/apps/web/src/store/SPEC.md
+++ b/apps/web/src/store/SPEC.md
@@ -75,7 +75,7 @@ per-workspace views/attention, terminal catalogs, and one **per-session chat run
   leaving the client labelling and keying reads off a value the host no longer has; a project never fetched or an id absent from its list is a **no-op** — the next
   `workspace.list` reconciles; **`applyWorkspaceRemoved(projectId, id)`** is the **entire** removal
   reaction (`removeWorkspace` drops the row + `clearWorkspaceState` drops its
-  local view/attention/terminal maps and chat runtimes + recency drops the dead id,
+  local view/attention/terminal/activity maps and chat runtimes + recency drops the dead id,
   and **if it was this client's active workspace** → activate the most recently selected loaded workspace
   whose project remains open, even across projects; when none remains, `selectProject(projectId)` falls back
   to the removed workspace's Project Home; either active fallback gets the same neutral toast that reads right
@@ -191,7 +191,22 @@ per-workspace views/attention, terminal catalogs, and one **per-session chat run
   otherwise the client renders the reply twice (frozen failed partial + retried copy). Hydration applies
   the same presentation rule to the persisted copy (`chat/hydrate.ts` hides retried attempts — an
   errored assistant followed by another assistant before any user message), so live and reloaded clients
-  agree. Closed
+  agree.
+- **Workspace activity** (`activityByWorkspace`) is the host's cross-workspace agent-state signal, the one
+  thing here that describes chats **nobody has open** — the Projects rail's glyphs. Keyed workspace →
+  session → `ActivityStatus`, and **idle is absence at every level**: a retraction deletes the session key
+  and then the workspace key once it empties, so "quiet" is an empty map rather than a map full of nulls.
+  A status that did not move is not a state write at all (the reducer returns the identical object), which
+  is what keeps an always-mounted rail from re-rendering on every event of every session.
+  **`applySessionActivity`** folds one `session.activity` push; **`hydrateSessionActivity`** **replaces**
+  the whole map from the `session.activityList` snapshot — replacement, not merge, because a reconnect must
+  not leave a glyph behind for a session that settled while the socket was down. Both refuse removed
+  workspaces and tombstoned sessions, so a late push cannot resurrect a deleted chat's glyph.
+  The rollup is **not** stored: `selectWorkspaceActivity`/`selectProjectActivity` derive it on read with a
+  single shared precedence, **`failed` > `waiting` > `running` > `queued`** — a rare fault must never be
+  masked by routine work, and both are "needs you" anyway. Note this is deliberately *not* the host's
+  per-session derivation order (see `packages/server/src/agent/SPEC.md`): there the question is "what is
+  this one chat doing", here it is "which of several chats should this row speak for". Closed
   chats are reopenable: the workbench close command atomically removes local placement and invokes
   **`closeChatToHistory`**, which **keeps the runtime + host session alive**, records it in
   **`closedChatsByWorkspace`** (`ClosedChat[]`, per workspace, most-recent-first), and clears pending
@@ -207,7 +222,7 @@ per-workspace views/attention, terminal catalogs, and one **per-session chat run
   gaining local placement. **`deleteChat(workspaceId, sessionId)`** is the idempotent
   fold for both a confirmed local `session.delete` and the `session.deleted` broadcast: it atomically drops
   every tab the chat owns — its transcript, live plan page, and any dependent legacy document cache — plus
-  its history row/runtime + skill baseline, records a page-lifetime tombstone, removes queued opens for the
+  its history row/runtime + skill baseline + activity row, records a page-lifetime tombstone, removes queued opens for the
   chat or its dependent documents, and queues a resource-removal intent. The shell layout integration
   removes every matching chat placement and session-backed plan reference through its pure mutation path,
   then reconciles local attention in the same transition. Until then the tombstone renders no body, so a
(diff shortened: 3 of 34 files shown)


apps/web/src/transport/wireTransport.ts
```diff
@@ -5,20 +5,34 @@ import type {
...
+import { ACTIVITY_PROTOCOL_VERSION, WS_CHANNELS } from "@thinkrail/contracts";
 import { isConnectedGeneration, useAppStore } from "../store";
 import { createPiEventBatcher, shouldFlushPiEventsBefore } from "./piEventBatcher";
 import { WsTransport } from "./transport";
 
 let transport: WsTransport | null = null;
 
+function refreshSessionActivity(connectionGeneration: number): void {
+	const { protocolVersion } = useAppStore.getState();
+	if (protocolVersion === null || protocolVersion < ACTIVITY_PROTOCOL_VERSION) return;
```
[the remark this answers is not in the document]
rsolmano: Fixed in 05399f43 — good catch, and it contradicted the contracts spec too ("a UI shipped ahead of its host renders no glyphs").

The gate is no longer an early return: when the welcome reports a pre-`ACTIVITY_PROTOCOL_VERSION` host, the client now hydrates an empty snapshot, which the replace-not-merge semantics turn into a clear. That covers the v59 → v58 transition you described, since an older host can send neither a replacement snapshot nor a retraction.

Two follow-ons while I was in there:
- extracted `supportsSessionActivity(protocolVersion)` so the capability rule is testable rather than a bare inline comparison (`wireTransport.test.ts`);
- made `hydrateSessionActivity` a genuine no-op when the computed map matches the current one — otherwise clearing on every pre-v59 reconnect would hand the rail a fresh object each time and re-render it.

Covered by three new store tests (empty snapshot clears; unchanged snapshot keeps object identity, both populated and empty; hydration still notices a changed status, an added chat and a dropped one) plus the two capability tests. `store/SPEC.md` and `contracts/SPEC.md` now record why the gate clears instead of returning.

apps/web/src/panels/ProjectTree.tsx
```diff
@@ -263,6 +268,11 @@ export function ProjectTree() {
 								project={project}
 								isSelected={selectedProjectId === project.id}
 								isExpanded={isExpanded}
+								activity={
+									isExpanded
+										? null
+										: projectActivityRollup(activityByWorkspace, workspaces, project.id)
```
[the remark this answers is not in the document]
rsolmano: Confirmed and fixed in 0f183ff1 — this one was a real hole in the feature's headline promise. `ProjectTree` loads `workspace.list` only for expanded projects (`if (!expandedProjectIds[project.id] || workspaces[project.id]) continue`), so a collapsed project that had never been opened would have rolled up to `null` — precisely the "a session is running somewhere I haven't looked" case.

Took your first suggestion: **attribution now travels with the row.** `SessionActivity` and the push payload carry `projectId`, resolved host-side through a new `setActivityProjectResolver` seam (the same pattern as `setSkillAdmissionResolver`, so the agent module stays ignorant of projects; an unresolvable workspace publishes nothing, since the client's workspace-removal fold has already dropped its activity).

I preferred that over eagerly hydrating every open project's workspace list, which would add requests the rail doesn't otherwise need and would still race activity arriving before the lists land.

The store slice became `Record<workspaceId, { projectId, sessions }>`, so `projectActivityRollup` now derives from the activity map alone — no `workspaces` argument, no second slice. Covered by two new tests that pin exactly your scenario (a project rollup with **no** workspace list loaded, and a workspace re-attributed to another project not being counted under both). `contracts/SPEC.md`, `store/SPEC.md` and `agent/SPEC.md` record why attribution rides the wire rather than being derived client-side.

Re-verified after the change: full e2e (all 8 shards), 14/14 workspaces' unit tests, typecheck, lint, and check:spec-surface.

packages/server/src/agent/activity.ts
```diff
@@ -0,0 +1,39 @@
...
+		const view = views[i];
+		if (!view) continue;
+		if (view.role === "user") return false;
+		if (view.role === "assistant") return view.stopReason === "error";
+	}
+	return false;
+}
+
+function failed(inputs: ActivityInputs): boolean {
+	if (inputs.lastSettlement !== undefined) return inputs.lastSettlement?.stopReason === "error";
```
danyaberezun: `length` is an actionable chat failure, but this returns idle for live and persisted terminals. Classify `error` and `length` consistently and test both paths.

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -116,6 +126,51 @@ export function setSessionDeletedPublisher(fn: (payload: SessionDeletedPayload)
...
+function retractActivity(sessionId: string, workspaceId: string): void {
+	const projectId = activityProjectId(workspaceId);
+	if (projectId === null) return;
+	publishActivity({ sessionId, workspaceId, projectId, status: null });
+}
+
+export function listSessionActivity(): SessionActivity[] {
+	const rows: SessionActivity[] = [];
+	for (const [sessionId, entry] of sessions) {
+		if (isSessionDeleted(sessionId, entry.workspaceId)) continue;
```
danyaberezun, line 237: A reconnect during deletion snapshots this session as absent. If trashing fails, the tombstone is removed without republishing, while `publishedActivity` still suppresses the old status. Keep activity until commit or force a rollback publish; test this race.

apps/web/src/panels/ActivityGlyph.tsx
```diff
@@ -0,0 +1,85 @@
...
+export function ActivityGlyph({
+	status,
+	counts,
+}: {
+	status: ActivityStatus;
+	counts?: Partial<Record<ActivityStatus, number>>;
+}) {
+	const { Icon, label, className } = PRESENTATION[status];
+	const lines = counts ? activityBreakdown(counts) : [];
+	const detailed = lines.length > 1;
```
danyaberezun: This counts distinct statuses, not chats: `{ running: 2 }` still gets the generic label. Use total chat count so repeated statuses show the promised breakdown.

apps/web/src/panels/ProjectTree.tsx
```diff
@@ -414,6 +425,7 @@ function ProjectRow({
 					{project.name}
 				</span>
 			</button>
+			{activity && <ActivityGlyph status={activity.status} counts={activity.counts} />}
```
danyaberezun, line 429: The spec requires `data-activity` on activity rows, but collapsed project rows only render the glyph. Add the row attribute and cover the collapsed-project path.

packages/server/src/agent/SPEC.md
```diff
@@ -113,6 +113,63 @@ answer-injection path, and the **restart repair** that keeps re-opened transcrip
     destructive queue API returns text but drops image blocks; Pi's queue events remain authoritative for
     membership/order. The host projects only a conservative `hasImages` aggregate into summaries/events, so
     image bytes do not ride the ordinary read stream.
+- **Activity projection** (`activity.ts`) answers "what is happening in a workspace nobody has open?" —
```
danyaberezun: This top-level bullet terminates `Boundary > Owns`; the remaining `agentSessionManager` contract becomes activity prose. Keep this section inside `Owns`.

apps/web/src/store/SPEC.md
```diff
@@ -191,7 +191,31 @@ per-workspace views/attention, terminal catalogs, and one **per-session chat run
   otherwise the client renders the reply twice (frozen failed partial + retried copy). Hydration applies
   the same presentation rule to the persisted copy (`chat/hydrate.ts` hides retried attempts — an
   errored assistant followed by another assistant before any user message), so live and reloaded clients
-  agree. Closed
+  agree.
+- **Workspace activity** (`activityByWorkspace`) is the host's cross-workspace agent-state signal, the one
```
danyaberezun: This top-level bullet terminates the store's `Owns` item, nesting the remaining store contract under activity. Keep this section inside `Owns`.

danyaberezun (changes requested):
Fix the inline findings; also update `host/SPEC.md` and `transport/SPEC.md` for the new activity wiring.

packages/server/src/agent/activity.ts
rsolmano: Valid — fixed in e5cf52aa. `store/SPEC.md` already says a `length` stop "becomes an actionable truncation error \u2014 neither may become '\u2713 Done'", so the rail calling a truncated run idle contradicted the transcript beside it.

Both paths now consult one `FAILED_STOP_REASONS` set (`error`, `length`) via a single `failedStopReason()` helper, so the settlement path and the transcript-fallback path cannot classify differently by construction. Tested both directions and both paths: `error`/`length` \u2192 failed, and `stop`/`toolUse`/`aborted` \u2192 idle, asserted for settlement *and* trailing-assistant inputs.

apps/web/src/panels/ActivityGlyph.tsx
rsolmano: Valid \u2014 fixed in e5cf52aa. `detailed` was `lines.length > 1`, so `{ running: 2 }` produced one breakdown line and fell back to the generic label, dropping the count in exactly the single-status case. My own test only covered `{ running: 1 }`, which is why it slipped through.

Now gated on `activityChatCount(counts) > 1` \u2014 total chats, per your suggestion. Added tests that `{ running: 2 }` renders "2 chats working", that the helper sums across statuses, and that a single chat still keeps the plain label whichever status it is in.

packages/server/src/agent/agentSessionManager.ts
rsolmano: Valid \u2014 fixed in e5cf52aa. The rollback path now calls `syncSessionActivity(sessionId)` right after `deletedSessions.delete(sessionId)`, so the entry republishes its true status once the tombstone is gone.

Worth recording how I confirmed it, because my first attempt at a test was worthless: asserting `listSessionActivity()` after the rollback passes with **or** without the fix, since the snapshot recomputes from live entries and never reads `publishedActivity`. The divergence is on the *push* side only. The test now reproduces it exactly \u2014 prompt an errored run (`publishedActivity = "failed"`), start the delete, wait for trash to be entered, call `syncSessionActivity` to stand in for any event arriving inside the tombstone window (it publishes `null` and latches `publishedActivity = null`), then fail the trash. Without the fix the last published status stays `null`; with it, `"failed"`. I verified it fails on the unfixed code (`Expected "failed", Received null`) before keeping it.

It also asserts the snapshot behaviour you described \u2014 the session reads as absent for the duration of the tombstone window, and attributable again afterwards.

apps/web/src/panels/ProjectTree.tsx
rsolmano: Valid \u2014 fixed in e5cf52aa. The attribute genuinely was missing from the project row: I had written it in a multi-edit that failed atomically on an unrelated hunk, then only retried the workspace-row half. So the spec said rows carry `data-activity` while collapsed project rows shipped only the glyph.

`data-activity` is now on the project row too, with a new e2e case that collapses the project and asserts the row attribute, the glyph's `aria-label`, and that expanding again clears it (expanded projects deliberately carry no rollup \u2014 their workspace rows already say it). `panels/SPEC.md` now states explicitly that **both** row kinds carry it.

packages/server/src/agent/SPEC.md
rsolmano: Valid \u2014 fixed in e5cf52aa. The `Owns:` list uses `  - ` sub-bullets with 4-space continuation prose; my column-0 `- **Activity projection**` closed the list, so everything after it \u2014 including the rest of the `agentSessionManager` contract and the `Lifetime:` paragraph \u2014 re-read as siblings of `Owns` rather than members of it.

Re-nested the whole activity block (bullet \u2192 `  - `, its prose/table/sub-bullets \u2192 4-space) so it sits alongside the `agentSessionManager` sibling above it. Caught while fixing this that my first mechanical shift moved 472 lines rather than my ~55, i.e. it re-indented pre-existing content; I reverted and scoped it to my own added lines using the diff against the base.

apps/web/src/store/SPEC.md
rsolmano: Valid \u2014 fixed in e5cf52aa, same cause as the agent SPEC one. My top-level bullet split the store's `Owns:` item, nesting the remaining store contract under activity.

Re-nested as a `  - ` sub-bullet inside `Owns:`. The mechanical shift also stranded the word `Closed` on its own line \u2014 my block had been spliced mid-sentence into "Closed chats are reopenable\u2026" \u2014 so that sentence is rejoined at the original indent.

packages/server/src/host/server.ts
```diff
@@ -344,6 +350,14 @@ export async function createServer(options: CreateServerOptions = {}): Promise<R
 		}
 	});
 
+	setActivityProjectResolver((workspaceId) => {
```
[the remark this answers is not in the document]
rsolmano: Valid \u2014 fixed in a285031f (rebased). Both modules gained real wiring in this PR and neither spec recorded it, which is exactly the boundary-erosion risk you describe: `check:boundaries` stays green because it checks import edges, not whether a spec owns them.

**`packages/server/src/host/SPEC.md`** \u2014 new *Activity fan-out* bullet, a sibling of the session-deletion fan-out: installs the agent's activity publisher, broadcasts `SessionActivityPayload` on `session.activity`, subscribes it for every client in the WS `open` handler, serves the `session.activityList` snapshot (and why one exists at all \u2014 pushes are never replayed), and satisfies the agent's `setActivityProjectResolver` seam from the workspace registry. It states that derivation/precedence/lifetime stay with [[submodule-server-agent]] and the wire shape with [[module-contracts]], so the boundary reads in one direction.

**`apps/web/src/transport/SPEC.md`** \u2014 `session.activity` \u2192 `applySessionActivity` added to the push-routing list, plus a paragraph on hydrate-on-welcome: `session.activityList` \u2192 `hydrateSessionActivity` under the connection-generation fence, and that failing `supportsSessionActivity` hydrates `[]` rather than skipping, so an older host clears glyphs instead of stranding them. Allowed-deps now names `SessionActivityPayload` + `ACTIVITY_PROTOCOL_VERSION`, and the public-surface bullet records that `supportsSessionActivity` stays module-internal.

Note the rebase onto #415 also caught a collision worth flagging: `JBCENTRAL_QUOTA_PROTOCOL_VERSION` had taken 59, and git auto-merged both features onto the same number. Activity is now **60** with `PROTOCOL_VERSION` at 60, both quota and activity assertions kept. Re-verified on the new base: full e2e (8/8 shards), 14/14 workspaces, typecheck, lint, and all four checks.

apps/web/src/transport/wireTransport.ts
```diff
@@ -5,20 +5,41 @@ import type {
...
+	if (!supportsSessionActivity(state.protocolVersion)) {
+		state.hydrateSessionActivity([]);
+		return;
+	}
+	void getTransport()
+		.request("session.activityList", {})
+		.then((rows) => {
+			const current = useAppStore.getState();
+			if (!isConnectedGeneration(current, connectionGeneration)) return;
+			current.hydrateSessionActivity(rows);
```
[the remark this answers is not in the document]
rsolmano: Valid \u2014 fixed in aa7b940b. You are right that the generation fence is the wrong instrument: the collision is *within* one connection, so it never trips.

Fixed as hydrate-then-stream rather than a revision fence, because a revision guard would have to **discard** the snapshot on collision, and that loses rows for every session the push did not touch \u2014 on a fresh page those glyphs would simply be missing. Instead `activityHydration` buffers `session.activity` for the duration of the read and replays it, in arrival order, immediately after the snapshot installs. The push therefore always lands last and wins, with no reconciliation bookkeeping in the store. It is the same shape the Pi-event batcher already uses in this module, and it keeps ordering in the transport (which owns push routing) with the store as a plain fold.

Three edges are handled deliberately, and each has a test:
- reads are **tokenized**, so a stale response cannot settle over a newer one;
- a **failed** read still replays its buffer \u2014 no snapshot arrived, so those pushes are the only truth left;
- a **superseded generation** discards the buffer instead, since a fresh welcome is already re-reading and replaying a dead connection's pushes would resurrect stale rows.

7 new tests in `activityHydration.test.ts` cover pass-through when idle, the exact clobber scenario you describe, replay ordering, and those three edges. `transport/SPEC.md` records why the generation fence is insufficient. Re-verified: full e2e (8/8 shards), 14/14 workspaces, typecheck, lint, check:spec-surface.

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -116,6 +126,51 @@ export function setSessionDeletedPublisher(fn: (payload: SessionDeletedPayload)
...
+
+function retractActivity(sessionId: string, workspaceId: string): void {
+	const projectId = activityProjectId(workspaceId);
+	if (projectId === null) return;
+	publishActivity({ sessionId, workspaceId, projectId, status: null });
+}
+
+export function listSessionActivity(): SessionActivity[] {
+	const rows: SessionActivity[] = [];
+	for (const [sessionId, entry] of sessions) {
```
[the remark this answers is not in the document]
rsolmano: Not disputing the analysis \u2014 but flagging that this is a **recorded, deliberate V1 boundary**, not an oversight, so I am getting a scope call before expanding the PR.

It is documented in three places already: `agent/SPEC.md` ("What no status survives is a session this process has never loaded: `listSessionActivity` walks live entries only\u2026 Extending the snapshot to derive those two from the disk-session scan (`listSessionInfosStrict`) is the deliberate follow-up, not V1"), plus the same limitation in the store spec and the PR description.

Where I think you have the stronger argument than the docs admit: **architecture.md decision #8** promises that "a reload, second tab, phone, or **host restart** therefore rebuilds the same projects, workspaces, sessions, and transcripts", and `session.list` already "unions in-memory sessions with pi's on-disk sessions". That makes `session.activityList` the one host read that *doesn't* union, which is a consistency argument I did not weigh when I scoped it out.

Against doing it in this PR: it needs a disk scan across every registered workspace plus a transcript-tail read per session on each snapshot, so it changes the cost profile of a connect (order of hundreds of tail reads for a large setup) and wants its own caching/bounding design and tests. That is a scope expansion on a design that was approved with this limitation named.

Deferring to the repo owner on whether to land it here or as the documented follow-up; I will update this thread either way.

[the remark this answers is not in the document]
rsolmano: Landed in 1b615d45 \u2014 the owner's call was to close it here, and the analysis of *where to store state* is worth recording since it shaped the fix.

**We do not store it.** I considered both obvious options and rejected them:
- **Metadata in the session file** \u2014 that file is pi's format, written by pi's `SessionManager`; adding our fields forks it and breaks plain-`pi` portability. The one sanctioned way in (`sendCustomMessage`, as `ask-user-answers` uses) appends a *message*, so status would enter the conversation and pollute the agent's context.
- **A status sidecar** (the shape `todos/<sessionId>.reviews.json` already uses) \u2014 mechanically idiomatic, but it would only ever cache a derivation, and it has a second writer: sessions live in pi's cwd-keyed directory, so plain `pi` writes the same transcripts. Answer the question in the terminal and the sidecar still says "waiting for your answer" \u2014 a *wrong* signal, strictly worse than a missing one. It would also need to stat every transcript to validate, saving only the tail read, and would churn writes on every running\u2194idle transition for data that is worthless after a restart.

Note `running` and `queued` are unpersistable in principle: both describe a live process, and the host can crash, so a stored value is a permanent lie.

**What landed instead:** `listSessionActivity` unions live entries with on-disk sessions per workspace, and disk rows go through `deriveDiskActivityStatus`, which pins `isStreaming: false`, `pendingMessageCount: 0`, `hasPendingDialog: false` and an `undefined` settlement \u2014 so `running`/`queued` are *structurally unreachable* from disk rather than filtered afterwards, and there is still exactly one precedence order. Both durable signals live at the transcript **tail** by construction, so a bounded 64 KiB window suffices (partial leading line dropped), memoized per file on `(mtime, messageCount)` **in memory only** so a fresh process re-derives.

Cheaper than I first estimated, too: `session.list` already opens every `.jsonl` to validate its header, so the walk and a file-open per session were already the cost of an existing read.

The handler passes `{id, cwd}` from `listAllWorkspaceRecords()`, keeping the agent's "cwd is an input, never a persistence lookup" boundary intact. An unreadable workspace is logged and skipped, never fatal.

Tests: 4 new pure ones (tail parsing incl. the partial-line and junk cases; disk rows can only be waiting/failed/idle; the disk path equals the live derivation), 2 integration (a seeded transcript surfaces with no session ever attached, and flips to idle when a later assistant lands \u2014 proving memo invalidation; an unresolvable project yields nothing), and a new e2e that reloads after seeding and asserts the glyph appears **while the workspace is never activated and no chat tab exists** \u2014 your exact restart scenario. The limitation prose is deleted from the specs and replaced with the union behaviour + why the transcript stays the only store.

Full gate re-run: e2e 8/8 shards, 14/14 workspaces, typecheck, lint, spec-surface, boundaries.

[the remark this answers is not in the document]
rsolmano: Valid \u2014 fixed in b0409730. And the failure is worse than described, which changed the shape of the fix.

Your case: a decisive record larger than the window means the read starts inside it, the fragment is dropped, and the run reads idle. Confirmed.

The one I found while writing the test: **`[huge assistant carrying the ask call][small ack]`**. The window catches a fragment of the assistant plus the *complete* ack, so the parse is **non-empty** \u2014 it just no longer contains the tool call. So the obvious mitigation ("grow when the parse comes back empty") silently misses `waiting`, and only boundary alignment is correct.

`readTranscriptTail` now probes the byte immediately before the window and grows \u00d78 (to `TRANSCRIPT_TAIL_MAX_BYTES` = 8 MiB) until that byte is a newline, so the window always begins at a record boundary and nothing is ever fragmented. Only a single record beyond the cap degrades to the old best-effort drop. Probing the byte rather than searching the decoded text also sidesteps a UTF-8 hazard: slicing bytes can split a multi-byte character, and the alignment check has to be exact.

Two tests, both verified to **fail on the previous code** before I kept them (a 200 KB terminal assistant with `stopReason: error` \u2192 `failed`; a 200 KB questionnaire followed by its ack \u2192 `waiting`). `agent/SPEC.md` records why dropping the fragment and retry-when-empty are both insufficient.

Also corrected while in there: those fixtures had `text: ""` alongside `content`, which `FixtureMessage` forbids (`{text} | {content}`, never both) \u2014 it typechecked only because the e2e directory sits outside the workspaces. Fixed in both the server test and the e2e spec.

Full gate re-run: e2e 8/8 shards, 14/14 workspaces, typecheck, lint.

apps/web/src/transport/wireTransport.ts
```diff
@@ -5,20 +5,50 @@ import type {
...
+
+const activityHydration = createActivityHydration({
+	apply: (payload) => useAppStore.getState().applySessionActivity(payload),
+	hydrate: (rows) => useAppStore.getState().hydrateSessionActivity(rows),
+});
+
+function refreshSessionActivity(connectionGeneration: number): void {
+	const state = useAppStore.getState();
+	if (!supportsSessionActivity(state.protocolVersion)) {
+		state.hydrateSessionActivity([]);
```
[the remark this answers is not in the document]
rsolmano: Valid on both halves \u2014 fixed in ce93df1f. The mechanism you identified is the one I had missed: **`WsTransport` resends pending requests across a reconnect**, so an outcome can arrive after the connection that issued it is gone.

Two changes:
- the unsupported-welcome path now **abandons** any in-flight hydration *before* clearing. `abandon()` invalidates the read's token, so the resent request's rejection \u2014 or a late success \u2014 is inert and cannot replay v60-era pushes after the clear.
- **both** outcomes are generation-fenced now, not just success: `.catch` discards instead of draining when the generation moved. Draining is correct only while the same connection is still live, where no snapshot arrived and the buffered pushes really are the only truth left.

Ordering is the load-bearing part, so the spec now says so explicitly rather than leaving it implicit in the code.

3 new tests: abandon empties the buffer and makes a later `fail`/`settle` on the old token a no-op; pushes apply straight through afterwards; and your exact sequence \u2014 buffer a push, abandon, then have the resent request rejected \u2014 asserting nothing replays after the clear.

Full gate: e2e 8/8 shards, 14/14 workspaces, typecheck, lint.

danyaberezun (approved):

PR #417, opened 2026-09-04
Start work dialog reworked

### LINKED ISSUE #420: Create workspaces with custom name
## Problem

The 'Create workspace' dialog lets you pick a project, a base branch, a model and a first prompt — but not a name. Every workspace is born as `workspace-N` and only gets a real name later, inferred from your first request (naive pass on the committed
prompt, refined by the agentic pass at `agent_settled`).

That is a good default when you don't know yet what the workspace is for. It is the wrong default when you do:

- **Renaming of workspace does not fix the branch name.** The `workspace.rename` wire method deliberately runs with `renameBranch: false`. The git branch stays `workspace-3` for the life of the worktree, and that branch is what you see in the eventual PR unless an AI agent decides to move worktree to better branch name. A name supplied *at create time* is the only path that also produces a matching branch slug.
- **Creating without a prompt leaves a permanent placeholder.** An empty prompt is a supported way to submit the dialog (you land in a ready composer), and there is nothing for auto-naming to read — the workspace stays `workspace-N` until you rename it by hand, with the branch stuck regardless.

## Proposal

An optional **Name** field in `NewWorkspaceDialog`, worktree mode only (hidden in Project-folder mode, which
creates nothing). 

---

🤖 Claude (via Claude Code) drafted this issue.

CommanderTvis:
## What changed

**The dialog no longer renames itself under the user.** Picking Isolated workspace / Project folder used to rewrite both the title and the description below it. The title is now the constant **Start work**, the target segment sits directly under it, and the only mode-aware prose is one line stating the difference:

- Isolated workspace — "A separate git worktree on its own new branch."
- Project folder — "Your project folder itself. No isolation, work lands on the current branch."

**Folder mode now says where the work lands.** Worktree mode offers a base-branch picker; folder mode had an empty slot and never named the branch. `git.listBranches` now also reports the repo's checked-out branch (`current`), and folder mode renders it as a plain `On <branch>` read-out — text, not a control, since there is nothing to pick.

**Both bottom-row triggers name what they set** (`MODEL` / `EFFORT`). The chat composer opts out of the labels, where that row competes with the prompt for width.

**Worktree mode gained a Name field** (closes #420), prefilled from a new `workspace.suggestName` request
with the host's next free `workspace-N`. Untouched it stays a placeholder — `create` omits it, the host
allocates the slot, and its prompt-driven auto-rename still applies. Edited, the typed name is sent and locks that rename out, so the branch matches the name from the start.

The skill caption was reworded in plainer language.

## Verification

`lint`, `typecheck`, `test`, `check:deps`, `check:boundaries`, `check:spec-surface`, and the full `bun run e2e` (6 shards, all passed).

## Screenshots

| | Before | After |
| --- | --- | --- |
| **Isolated workspace** | before-worktree | after-worktree |
| **Project folder** | before-folder | after-folder |

### DIFF
--- a/apps/web/src/chat/Composer.tsx
+++ b/apps/web/src/chat/Composer.tsx
@@ -676,6 +676,7 @@ export const Composer = forwardRef<ComposerHandle, ComposerProps>(function Compo
 							refreshing={modelsRefreshing}
 							onRefresh={onRefreshModels}
 							onSelect={onSelectModel}
+							showLabel={false}
 							className="max-w-80 gap-4 px-4 sm:max-w-144"
 						/>
 						<ThinkingSelector
--- a/apps/web/src/chat/ModelSelector.tsx
+++ b/apps/web/src/chat/ModelSelector.tsx
@@ -39,6 +39,7 @@ export function ModelSelector({
 	placeholder,
 	defaultOption,
 	onSelectDefault,
+	showLabel = true,
 }: {
 	models: WireModel[];
 	current: WireModel | null;
@@ -50,6 +51,7 @@ export function ModelSelector({
 	placeholder?: string;
 	defaultOption?: string;
 	onSelectDefault?: () => void;
+	showLabel?: boolean;
 }) {
 	const [open, setOpen] = useState(false);
 	const providers = [...new Set(models.map((m) => m.provider))];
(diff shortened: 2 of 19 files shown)


CommanderTvis:
non-priority, closing to reduce team's review load

PR #430, opened 2026-09-05
fix(server): thaw a terminal frozen by a drain that never arrived

### LINKED ISSUE #428: Terminal freezes after macOS sleep: a lost WebSocket drain event latches backpressure forever
## Symptom

After a macOS hibernate (in my case combined with a monitor/App move), a terminal tab stops updating entirely while the process underneath keeps running — Claude Code kept working in the pty, but the pane showed a frozen frame. The WebSocket between webview and host was still `ESTABLISHED` (verified with `lsof`), so it isn't a dead connection. Reloading the window recovers, which is why this hides well.

## Root cause

Terminal output flow control is edge-triggered with no recovery path:

1. During sleep the webview stops reading its socket, a `ws.send` on the host backpressures, and the host latches the client in `terminalBackpressured` while the tab's output batcher sets `blocked`.
2. The **only** thing that lifts the latch is Bun's `drain` event (`packages/server/src/host/server.ts`). The batcher deliberately retries only on `resume()` — that's a pinned semantic — so the latch's lift must be guaranteed.
3. A drain lost across a system sleep is never re-delivered. The latch stays set forever: every later flush returns "unavailable", the tab freezes, the pty and the output recorder stay live underneath.

## Fix

Don't trust the drain event alone — the socket itself knows the truth. A 1-second reconciler checks each latched client: `getBufferedAmount() === 0` with the flag still set means the OS never delivered the drain, so the host fires it (clears the flag, resumes the client's terminals). Real backpressure is untouched — a socket with buffered bytes stays latched until drain or the reconciler sees it empty.

Fix on my branch:

CommanderTvis/thinkrail@efa2f00de5b30938c5b50ba5db02c6426df9ff29

Happy to send it as a standalone PR.

---

CommanderTvis:
## Summary

After a macOS hibernate a terminal tab stops updating while its pty keeps running. The WebSocket is still `ESTABLISHED`; only a window reload recovers it.

Terminal output flow control is edge-triggered with no recovery path. A backpressured `ws.send` latches the client in `terminalBackpressured` and the tab's output batcher sets `blocked`; the batcher deliberately retries only on `resume()`, so the latch's lift has to be guaranteed. The only thing that lifted it was Bun's `drain` event — and a drain lost across a system sleep is never re-delivered. Every later flush then returns `unavailable`, forever.

The fix stops trusting the drain event alone, because the socket itself knows the truth. A 1s reconciler checks each latched client, and `getBufferedAmount() === 0` with the flag still set means the OS never delivered the drain — so the host does what the drain would have: clears the flag and resumes that client's terminals.

Real backpressure is untouched. A socket with buffered bytes stays latched until drain or until the reconciler sees it empty. A latch whose socket is already gone is left to `close`/`open`, which own that case.

The reconcile predicate lives in `packages/server/src/host/terminalSend.ts` as `drainedClientKeys`, next to `terminalDeliveryForSendStatus`, so it is unit-tested rather than buried inside `createServer`. The interval is cleared in `stop()`.

## Related issues

Closes #428

## Checklist

### On the E2E box

Re-run after rebasing onto `main`: `bun run e2e` is fully green across all 6 shards (393/393 passed in 240s). The earlier flakiness and failures on `main` (`projects.spec.ts:277`, `layout.spec.ts:411`, `welcome.spec.ts:32`, `chat-order.spec.ts:117`) have all cleared.

Fast gates (`check:deps`, `check:boundaries`, `check:seams`, `lint`, `typecheck`) are green.

The post-sleep recovery itself is not reachable from the suite — it needs a genuinely backpressured socket plus a suspended host — so there is no automated coverage of the end-to-end thaw; the reconcile decision is covered by the unit tests in `terminalSend.test.ts`.


danyaberezun (changes requested):
Rebase needed: conflicts with main (post-#467) in packages/server/src/host/server.ts — both PRs add declarations right after `terminalBackpressured = new Set...`. Purely textual, keep both blocks.

Fix itself checks out: diagnosis matches the code (edge-triggered drain, batcher only retries on resume()), getBufferedAmount()===0 is the correct level-triggered fallback, interval lifecycle mirrors reapTimers/hostUpdateTimer, no races (single-threaded, idempotent resume). Lint/typecheck/tests pass on the branch.

Optional, non-blocking: `drain(ws)` and the reconciler both run `terminalBackpressured.delete(clientKey); resumeClientTerminals(clientKey);` — could factor into one `liftBackpressureLatch(clientKey)` helper.

CommanderTvis:
non-priority, closing to reduce team's review load

PR #454, opened 2026-09-09
Focus tab issue

OLavrik:
Cleaned up focus-ring rendering across the shell. TreeRow (file tree) now has a unified inset focus-visible ring matching other lists (committed 63ae5d44). The SpecsPanel single-ring-on-focus behavior was found already present in HEAD, so no change was needed there. This is on top of the earlier tab focus-ring alignment fix (04125dee). Changes are CSS-class-only; not run through e2e since deps aren't installed in this worktree.

### Clean up focus-ring rendering
- [x] **SpecsPanel: suppress active-row ring when focused (single ring)** (`c102d33`)
  No-op: SpecsPanel already carries has-[:focus-visible]:ring-0 on the active row in HEAD, so an active+focused row already collapses to a single focus ring. No diff produced.
  Verified: Confirmed via git show HEAD — the class was already present; git diff for SpecsPanel is empty.
- [x] **TreeRow: add unified focus-visible ring**
  Added outline-none + focus-visible:ring-2 ring-inset ring-primary to the TreeRow button (h-24 w-full rounded-sm), so file/dir tree rows now show the same clean inset focus ring as other lists instead of the default browser outline.
  Verified: not verified (CSS-class-only; deps not installed in worktree). Committed as 63ae5d44.
Screenshot 2026-09-09 at 05 04 38

### DIFF
--- a/apps/web/src/panels/SpecsPanel.tsx
+++ b/apps/web/src/panels/SpecsPanel.tsx
@@ -137,7 +137,7 @@ function SpecNodeRow({
 				className={cn(
 					"group flex h-28 min-w-0 items-stretch rounded-[var(--radius-sm)] px-4 transition-colors",
 					isActive
-						? "bg-primary-subtle ring-1 ring-primary-muted ring-inset"
+						? "bg-primary-subtle ring-1 ring-primary-muted ring-inset has-[:focus-visible]:ring-0"
 						: "hover:bg-control-bg-hovered",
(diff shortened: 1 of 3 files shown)


danyaberezun:
As for me, i prefer old-style

rsolmano (approved):

danyaberezun (approved):

PR #460, opened 2026-09-09
Fixed the stuck queue (Steering mode)

OLavrik:
Fixed the stuck queue chip when an image is sent mid-run and hardened it across all submit modes. Root cause is a pi defect: it clears a delivered queued message by text-match, so an image-only (empty-text) steer or follow-up stays in pi's queue forever. The host workaround in agentSessionManager counts such stuck deliveries per lane, synthesizes the withheld queue_update, and drops the phantom empties from every display read (queue_update projection, mirror sync, queueStateOf, effective pendingMessageCount). Audited the four submit modes: steer and followUp both hit the defect and are both covered by the fix; send/interrupt are un-queued prompts with no phantom. Added three tests (steer-lane clear, follow-up-lane clear, idle-image-prompt counter-safety) and updated the agent SPEC. Verified: server bun test 209 agent / 988 total pass, typecheck + biome + boundary/catalog/seam gates green.

### Clear stuck image-only Steering chip
- [x] **Confirm queue_update/PiEvent + user content types in contracts** (`d47bc14`)
  Confirmed PiEvent.queue_update shape (steering/followUp readonly string[], optional hasImages) and that user message content blocks use {type:'image'|'text'}.
  Verified: Read contracts/src/piProtocol.ts and existing image-steer tests.
- [x] **Update packages/server/src/agent/SPEC.md with the phantom-clear workaround**
  Documented the image-only steer/follow-up phantom-clear workaround (stuckEmptyDeliveries per-lane counter, synthesized queue_update, drop-empties on all display reads) in the queue-projection section of the agent SPEC.
  Verified: Re-read the edited SPEC section.
- [x] **Implement host workaround in agentSessionManager** (`1f0d85f`)
  Added per-lane stuckEmptyDeliveries counter to Entry. On a user message_start with empty text + image content matching a pending empty-lane entry, the manager increments the counter and synthesizes the queue_update pi withheld. displayedLane() drops that many empty entries from pi's list and is now the single source for the projected queue_update, mirror sync, and queueStateOf; clearQueueSession resets the counter.
  Verified: bunx tsc --noEmit in packages/server — clean.
- [x] **Add unit test for image-only steer delivery clearing the chip** (`c6c4820`)
  Added a gated tool-call test that steers an image with empty text mid-run, asserts the queued chip shows, then after delivery asserts summary.queue is undefined, activity is not 'queued', and the final queue_update carries steering:[].
  Verified: bun test agentSessionManager.test.ts — 49 pass, 0 fail.
- [x] **Run agent unit tests + typecheck/lint** (`d82a89b`)
  Ran gates on the change: biome clean (after autoformat), tsc --noEmit clean for packages/server, and the full server bun test suite passes.
  Verified: biome check clean; tsc --noEmit clean; bun test (server) — 988 pass, 0 fail.
- [x] **Cover follow-up lane + un-queued-mode safety**
  Audited all four submit modes: steer (steering lane) and followUp (follow-up lane) both hit pi's empty-text defect and are both covered by the host fix; send/interrupt are un-queued prompts and produce no phantom. Added a follow-up-lane delivery-clears test and an idle-image-only-prompt guard test proving the counter isn't falsely incremented (a later real image-only steer still shows as queued and clears).
  Verified: bun test src/agent — 209 pass, 0 fail; biome clean; comment audit empty.

Review: 0/6 steps reviewed in ThinkRail.


danyaberezun (approved):

rsolmano (changes requested):
Looks like we're fixing a symptom here. Please review the fix in the Pi first
https://github.com/earendil-works/pi/pull/8612

rsolmano (approved):

PR #482, opened 2026-09-13
TODO queue bug vs ask user question

OLavrik:
Fixed the bug where a message queued while the agent was working would supersede a pending ask_user_question (the agent jumped to the queued task instead of waiting for the answer). Diagnosed it as a consequence of our host-owned ack+terminate ask design (not a pi bug — pi's own blocking dialogs never drain the queue), and confirmed the fix belongs in the host with the user (Approach A; Skip flushes held messages in order). Implemented a per-session QuestionHold seam: the ask tool synchronously drains pi's queue into entry.heldWhileAsking before terminating (race-free), all send paths park while a question is pending, the held buffer is unioned into every queue projection but excluded from pendingMessageCount (so status stays "waiting", not "queued"), and answerQuestion (answer or Skip) flushes held messages in order after the reply. Updated packages/server/src/agent/SPEC.md and added an integration test. Verified: typecheck 15/15, agent unit tests 207 pass, lint clean, full e2e gate all 8 shards green (one unrelated flake self-resolved).

### Fix: queued message shouldn't supersede a pending question
- [x] **Confirm scenario + desired behavior with user**
  User confirmed scenario A (queued followUp superseded a pending ask_user_question) and chose: the pending question should hold the queue until answered/dismissed.
  Verified: not verified (clarification only)
- [x] **Brainstorm the fix (task-spec) against agent/activity specs**
  Diagnosed root cause (our ack+terminate design, not a pi bug) and recorded Approach A + Skip-flush decision in .thinkrail/context/TASK-question-holds-queue.md with a concrete host-side implementation plan.
  Verified: not verified (design only)
- [x] **QuestionHold seam: hold-aware ask tool + buildResourceLoader wiring** (`3785165`)
  Added a QuestionHold seam: createAskUserQuestionTool(hold?)/askUserQuestionExtensionFor(hold?), buildResourceLoader threads an optional questionHold, and the two host-session creation paths bind it to the entry.
  Verified: bun run typecheck green; agent unit tests pass
- [x] **Capture on question + hold sends + flush on answer in manager**
  Tool execute synchronously captures+clears pi's queue into entry.heldWhileAsking when a question is raised; prompt/steer/followUp park while a question is pending; answerQuestion flushes held in order after the answer.
  Verified: new integration test in agentSessionManager.test.ts passes (hold→waiting→flush)
- [x] **Union held buffer into queue projection + clear/abort integration**
  Held buffer unioned into queueContentOf/queueStateOf/hasQueuedImages and the queue_update projection (heldQueueEvent); summaryOf queue gate includes held; clearQueueSession clears held (so abort-restore stays lossless). Activity stays waiting because held is not in pi's pendingMessageCount.
  Verified: agent unit tests pass (49) incl. queue/abort/remove cases
- [x] **Update agent SPEC + tests, run agent unit tests** (`171a75d`)
  Updated agent/SPEC.md: rewrote the activity-precedence rationale (a pending question now holds the queue, so queued never co-occurs with waiting) and the askUserQuestion block (QuestionHold seam, capture/park/flush, projection union), plus the buildResourceLoader signature. Added an integration test covering hold→waiting→ordered flush.
  Verified: bun run typecheck green (15/15); bun test packages/server/src/agent 207 pass; bun run lint clean
- [x] **Full e2e gate**
  Ran the full e2e gate. First run had one flaky failure (chat-order.spec scroll ordering, unrelated to this change); it passed on --last-failed and the full re-run passed all 8 shards.
  Verified: bun run e2e — all 8 shards passed in 177.3s

Review: 1/6 steps reviewed in ThinkRail.

### DIFF
--- a/packages/server/src/agent/SPEC.md
+++ b/packages/server/src/agent/SPEC.md
@@ -468,10 +468,23 @@ answer-injection path, and the **restart repair** that keeps re-opened transcrip
     race-free; **(b)** while a question is pending (derived from the transcript via
     `awaitingQuestionToolCallId`, the single authority), `promptSession`/`steerSession`/`followUpSession`
     park into the same buffer instead of touching pi. The buffer is unioned into every queue projection
-    (`queueContentOf`/`queueStateOf`/`hasQueuedImages`/the `queue_update` fan-out and the `summaryOf`
-    queue field) so held messages still render as queued chips; it is **not** counted in
-    `pendingMessageCount`, which is why the status stays `waiting` rather than `queued`. `clearQueue`
-    empties it too, so Stop/`abortSession(..., true)` stay lossless. Pinned by `agentSessionManager.test.ts`.
+    (`queueContentOf`/`queueStateOf`/`hasQueuedImages`/`queueUpdateEventOf` and the `summaryOf` queue
+    field) so held messages still render as queued chips; it is **not** counted in `pendingMessageCount`
+    (nor `effectivePendingCount`), which is why the status stays `waiting` rather than `queued`. Capture
+    also **resets `stuckEmptyDeliveries`** since it empties pi's lanes (the same invariant
+    `clearQueueSession` holds), and it always clears — even when only stuck-empty placeholders are queued —
+    so no phantom entry drains past the boundary. `clearQueueSession` empties the buffer too, so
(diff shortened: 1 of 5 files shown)


rsolmano:
@OLavrik should we close it? 
actual bug should be fixed already in https://github.com/JetBrains/thinkrail/pull/517

PR #510, opened 2026-09-17
feat: unify attention and show live workspace activity

rsolmano:
## Summary

- Replaces the Projects rail’s multi-state activity vocabulary with one durable, owner-global “Needs attention” signal for completed, failed, length-limited, blocked, and host-interrupted chats.
- Clears review attention only after the exact result is genuinely rendered in a focused, unobscured conversation; unresolved blockers remain marked, while queued work and explicit user cancellation stay quiet.
- Adds a separate ambient running treatment: existing chat, workspace, and collapsed-project icons pulse without changing colour while a user-visible top-level chat is active. Reconnect restores live state; internal reviewer/reflector sessions remain hidden from both signals.
- Removes the old activity derivation/UI while retaining the deprecated empty `session.activityList` compatibility tombstone for old clients.

Verification:
- Fast gates, spec-surface validation, and the full unit suite pass (`1095` server tests passed, `2` platform skips).
- Full no-agent E2E passes: `401` tests across 8 shards.
- Focused real-agent restart/reconnect/question coverage passes: `4` tests.
- The full real-agent sweep was attempted twice. Across retries, 49/52 pass; the remaining failures are in unchanged chat-scroll runway and phone composer-width assertions.

## Checklist


packages/server/src/agent/agentSessionManager.ts
```diff
@@ -238,90 +197,481 @@ async function readTranscriptTail(path: string): Promise<AgentMessage[]> {
...
-		if (sessions.has(info.id) || isSessionDeleted(info.id, workspaceId)) continue;
-		rows.push({
-			sessionId: info.id,
-			status: await diskActivityStatus(info),
-			recencyMs: info.modified.getTime(),
-		});
+	const rows: WorkspaceAttentionRow[] = [];
+	for (const file of await scanSessionFiles(cwd, liveFiles)) {
+		if (!file.ok) {
+			log.warn(`attention ${mode} skipped transcript ${file.path}`, file.error);
```
danyaberezun, line 374: Do not return a partial scan as authoritative. This omission, and the catch below, can drop a previously known disk candidate; the client then replaces its entire attention map, and no unchanged-candidate push repairs it. Return completeness and preserve incomplete scopes, or fail the snapshot.

apps/web/src/transport/wireTransport.ts
```diff
@@ -6,50 +6,89 @@ import type {
...
 	void getTransport()
-		.request("session.activityList", {})
+		.request("session.attentionList", {})
 		.then((rows) => {
-			if (current()) activityHydration.settle(token, rows);
-			else activityHydration.discard(token);
+			if (current()) attentionHydration.settle(token, rows);
+			else attentionHydration.discard(token);
 		})
 		.catch(() => {
```
danyaberezun, line 67: Retry current-generation snapshot failures. Otherwise old attention rows remain visible but unacknowledgeable because their generation is stale; running was cleared, and an already-running session emits no new `true`. Both remain wrong until another reconnect.

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -238,90 +197,481 @@ async function readTranscriptTail(path: string): Promise<AgentMessage[]> {
...
+	if (
+		cached &&
+		cached.modifiedMs === metadata.mtimeMs &&
+		cached.size === metadata.size &&
+		(!requireFullRead || cached.fullRead)
+	) {
+		return cached.candidate;
+	}
+	const tail = await readAttentionTranscriptTail(path);
+	const fullRead = requireFullRead && tail.truncated;
```
danyaberezun, line 348: Modern transcripts never full-read here. A valid user JSONL record can exceed the 8 MiB tail—the composer permits 24 MiB of base64 images—so it is dropped as the partial first line. The final assistant then lacks its parent, derivation returns null, and that miss is cached. Detect incomplete ancestry before caching.

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -852,36 +1265,43 @@ async function openDiskSession(sessionId: string, workspaceId: string, cwd: stri
 		session.dispose();
 		return;
 	}
-	await registerSession(session, workspaceId, generation);
+	const userVisible = attachingUserVisible.get(sessionId);
```
danyaberezun, line 1268: This samples visibility before `registerSession()` awaits extension binding. A concurrent `userVisible: false` attach can update the shared map after this read, allowing a visible attention publication before the caller hides it. Keep the shared false decision authoritative through registration and publication.

apps/web/src/chat/useChatTodos.ts
```diff
@@ -238,6 +238,7 @@ async function nudgeAgent(workspaceId: string, sessionId: string, title: string)
 		});
 	} catch {
 		try {
+			const attentionReconciledEpoch = useAppStore.getState().attentionHydrationEpoch;
```
danyaberezun, line 241: This fallback is too late: the initial admission decision uses an optional client runtime. If another client has attached an idle session awaiting `ask_user_question`, `session.prompt` succeeds and supersedes that question. Use a host-side nudge operation that atomically chooses no-op, follow-up, or prompt from canonical state.

packages/server/src/agent/attention.ts
```diff
@@ -0,0 +1,233 @@
...
+	return `legacy-interrupted:${digest}`;
+}
+
+export function deriveAttentionCandidate(inputs: AttentionInputs): AttentionCandidate | null {
+	const turn = latestUserTurn(inputs.entries);
+	const turnId = turn?.entry.id ?? null;
+	if (inputs.pendingDialogId !== null) {
+		return { id: `dialog:${inputs.pendingDialogId}`, kind: "blocking", turnId };
+	}
+	const messages = attentionMessages(inputs.entries);
```
danyaberezun, line 121: Use one shared ask-lifecycle classifier. The host treats a non-ack tool result as resolved, while web `deriveAskStates` ignores tool results and can still report `waiting_question`; card, plan, attention, and nudge policy then disagree. Share the tool name and lifecycle truth table through contracts.

packages/server/src/agent/attention.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export interface AttentionInputs {
+	entries: readonly SessionEntry[];
+	isStreaming: boolean;
+	pendingMessageCount: number;
+	lastSettlement: AgentSettlement | null | undefined;
+	pendingDialogId: string | null;
+}
+
+const REVIEW_STOP_REASONS: ReadonlySet<string> = new Set(["stop", "error", "length"]);
```
danyaberezun, line 23: Do not add another stop-reason truth table. Web failure UI, analytics, reviewer cleanup, and delegation already classify these and have drifted: `pi-delegation` maps `length` to `completed`. Introduce one terminal-outcome classifier and derive purpose-specific projections from it.

danyaberezun (changes requested):
Verified against the current head. Also address these shared-state inconsistencies:

- `PlanPane.tsx:296` hardcodes `working`; `ReviewPanel.tsx:348` defaults to `waiting` even when the running map says otherwise. Use one selector.
- Capture `attentionHydrationEpoch` inside `getSessionMessagesWithSkillBaseline`; five call sites duplicate it before `watchReady`.
- Remove unused `SnapshotHydration.buffered()`.

PR #517, opened 2026-09-18
fix(agent): keep queued input behind live questions

rsolmano:
## Problem

A steering or follow-up message queued while `ask_user_question` was running could be consumed immediately after the tool's ack, superseding the unanswered questionnaire. The existing always-ack design avoided dangling tool calls across restart, but made Pi consider the live tool complete too early.

## Approach

Use a hybrid lifecycle on the current Pi APIs. Live questions block in a session-bound phase registry and return the user's real answer as the native tool result, so Pi naturally keeps its own queues behind the question. If the process restarts, attach-time repair closes an eligible dangling call with the existing ack and the later answer uses the existing custom-message continuation path.

The registry covers the full `expected → waiting → answer-accepted-uncommitted → complete` lifecycle. The first accepted terminal action wins: Stop claims an unanswered question before signalling Pi; an already accepted answer persists before Stop aborts only its continuation. Graceful shutdown freezes new answer admission, preserves unanswered questions for restart repair, and settles already accepted answers. Length-truncated assistant tool calls are never executable or answerable.

## Changes

- Block valid live `ask_user_question` calls until `session.answerQuestion` resolves them, and acknowledge Submit only after the native result is persisted.
- Treat `length`/`error`/`aborted` assistant calls as terminal through one shared executable-tool classifier; length-truncated restart repair writes an error rather than reviving a possibly incomplete questionnaire.
- Repair eligible dangling post-restart asks to an answerable ack; require that ack before using the custom-message fallback.
- Keep Pi steering/follow-up queues pending behind live questions; queued composer input no longer supersedes the card.
- Make expected, answer, Stop, shutdown, and disposal races explicit. Deferred Stop drains again immediately before abort so messages queued while an accepted answer persists are restored in lane order.
- Derive waiting/answered/stopped state from both native tool results and restart custom answers across activity, plan glance, questionnaire rendering, and dead-tool presentation.
- Update the owning specs and workflow guidance.

## Screenshots

Not included — PR creation was requested without additional screenshot capture; the interaction is covered by focused real-agent E2E.

## Checklist

- [ ] Before/after screenshots are included for frontend changes, or marked not applicable

## Testing

- `bun run test` — 1154 passed, 2 skipped, 0 failed.
- `bun run typecheck` — 15/15 packages passed.
- `bun run lint` — passed with five pre-existing unused-suppression warnings.
- `bun run check:deps`, `bun run check:boundaries`, `bun run check:seams`, `bun run check:spec-surface` — passed.
- `bun run e2e:agent -- e2e/ask-restart.live.spec.ts e2e/ask-user-question.live.spec.ts` — 10 passed.
- `bun run e2e:full` — 405/406 no-agent tests passed; unchanged `chat-scroll.spec.ts` touch-momentum coverage failed once and passed immediately in a focused serial rerun.


packages/server/src/agent/askUserQuestion.ts
```diff
@@ -287,32 +302,204 @@ export function buildAnswersMessage(
...
+		},
+		persistTurn(toolResults) {
+			for (const result of toolResults) {
+				if (result.toolName !== ASK_USER_QUESTION_TOOL_NAME) continue;
+				const waiter = waiting.get(result.toolCallId);
+				if (!waiter) continue;
+				waiting.delete(result.toolCallId);
+				waiter.cleanupAbort();
+				if (waiter.answer === undefined) continue;
+				if (waiter.waitStarted) waiter.resolvePersisted();
```
OLavrik: `persisted` only resolves when `turn_end` carries this tool result. In the answer+Stop race, if an abort makes the turn end without our just-returned result, this `waitStarted` waiter stays in the map and `persisted` never resolves, so `answerQuestion`'s `await live.persisted` hangs. Worth confirming Pi always includes a completed tool result in `turn_end.toolResults` even when aborted right after `execute` returns.

apps/web/src/chat/tools/AskUserQuestionCard.tsx
```diff
@@ -351,7 +341,7 @@ export function AskUserQuestionCard({
 	const cardRef = useRef<HTMLElement>(null);
 	const questions = useMemo(() => parseQuestions(args), [args]);
 	const resolvedResult = ask?.answer ?? readAskResult(result);
-	const awaiting = !resolvedResult && !ask?.superseded && status !== "error";
+	const awaiting = !resolvedResult && !ask?.superseded && !ask?.terminal && status !== "error";
```
OLavrik, line 344: There is no explicit `ask?.terminal` render branch — the dead/stopped visual relies on `ActivityGroup` mapping dead calls to `status:"error"`. The third terminal case (`done && no answer && !ack`) yields `status:"done"` and would fall through to the interactive card. Unreachable in practice, but an explicit `if (ask?.terminal) return <ResolvedRecord result={null} .../>` would remove the dependency on that convention.

rsolmano: Added the explicit terminal render branch in 0ab52196, so terminal done/no-answer states cannot fall through to the interactive card.

packages/server/src/agent/askUserQuestion.ts
rsolmano: Added a real Answer+Stop integration regression in 0ab52196. It confirms Pi includes the native result in `turn_end`, and both RPCs settle.

packages/server/src/agent/askUserQuestion.ts
```diff
@@ -287,32 +302,204 @@ export function buildAnswersMessage(
...
 			const validation = validateQuestionnaire(args);
 			if (!validation.ok) return toolResult(validation.message, { answers: [], cancelled: true });
 
-			return {
-				content: [{ type: "text", text: ASK_ACK_TEXT }],
-				details: { kind: "ack" } satisfies AskUserQuestionAckDetails,
-				terminate: true,
-			};
+			const outcome = await waiters.wait(toolCallId, signal);
+			if (outcome.kind === "abandoned") {
```
OLavrik, line 540: On graceful shutdown, `disposeAllSessions()` calls `abandon()` after `settleSessionsForShutdown()` deliberately left the question streaming. `abandon()` resolves the waiter with `{kind:"abandoned"}`, so this branch returns a terminal `cancelled` result. The "shutdown leaves the question recoverable" guarantee then relies entirely on `session.dispose()` (called synchronously right after `abandon()`) preventing this result from being persisted, so restart repair can write the `{kind:"ack"}` instead. It works against the faux SessionManager in the test, but it's an implicit ordering contract with real Pi: if Pi ever persists a tool result returned after `dispose()`, a gracefully-shut-down question is closed as `cancelled` (non-recoverable) rather than recoverable. Worth asserting this in the SPEC, or having `abandon()` not drive `execute` into the cancelled branch at all (leave the promise unresolved / no-op) since the result must be discarded regardless.

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -1402,7 +1446,8 @@ export function disposeAllSessions(): void {
 export async function settleSessionsForShutdown(timeoutMs = 2000): Promise<void> {
 	const settling = new Set<Promise<unknown>>();
 	for (const [sessionId, entry] of sessions) {
-		if (entry.session.isStreaming) settling.add(entry.session.abort());
+		if (entry.session.isStreaming && !entry.askUserQuestionWaiters.hasActiveCall())
```
OLavrik: This correctly preserves a live question on shutdown via `!hasActiveCall()`. Note the asymmetry: `removeWorkspaceSessionsInternal` aborts a streaming session unconditionally (no `hasActiveCall()` guard), so removing a workspace kills an unanswered live question with a stopped error. Fine for destructive removal, but worth confirming it's intentional — it's the one place the abort-guard differs from the shutdown path.

packages/server/src/agent/askUserQuestion.ts
OLavrik: Thanks for adding the regression in `0ab52196`. One ask: please confirm the test actually reproduces the hazard — i.e. it drives `turn_end` **without** this tool result after `execute` has already returned the answer, and asserts `answerQuestion` still settles. If the test calls `persistTurn` with the answered result included (the happy path), the exact race where `await live.persisted` can hang — abort makes the turn end without our just-returned result, leaving a `waitStarted` waiter stuck in the map — stays uncovered.

rsolmano: Confirmed both sides now. The gated Pi integration tests in `00b434a6`/`8a44ccf1` cover Answer winning before Stop, and `a0858fd9` adds the inverse terminal case explicitly: after execute returns the answer, `turn_end` without that result rejects the persistence promise instead of leaving it pending. The RPC therefore either acknowledges the persisted native result or fails; it cannot hang.

packages/server/src/agent/askUserQuestion.ts
```diff
@@ -287,32 +306,232 @@ export function buildAnswersMessage(
...
+		},
+		persistTurn(toolResults) {
+			for (const result of toolResults) {
+				if (result.toolName !== ASK_USER_QUESTION_TOOL_NAME) continue;
+				const waiter = waiting.get(result.toolCallId);
+				if (!waiter) continue;
+				waiting.delete(result.toolCallId);
+				waiter.cleanupAbort();
+				if (waiter.phase !== "answer-accepted-uncommitted") continue;
+				if (waiter.executeStarted) waiter.resolvePersisted();
```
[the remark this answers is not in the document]
rsolmano: Addressed in `a0858fd9`. The new semantic-validation regression accepts an answer during the expected phase, runs an invalid reserved-label questionnaire, and proves the answer persistence rejects rather than acknowledging the validation result. This is enforced by the existing `executeStarted` boundary: validation returns before `wait()`, so `persistTurn` cannot resolve the accepted answer.

packages/server/src/agent/askUserQuestion.ts
rsolmano: Documented the contract explicitly in `a0858fd9`: `abandon()` is disposal plumbing, not semantic cancellation, and is immediately followed by synchronous session disposal. The existing real-`SessionManager` shutdown/reopen test pins that no returned cancelled result lands before attach-time ack repair; a future Pi ordering change will fail that regression.

packages/server/src/agent/askUserQuestion.ts
rsolmano: Added the exact missing-result regression in `a0858fd9`: the tool executes and returns the accepted answer, then `persistTurn([])` simulates `turn_end` without that result and the answer persistence is asserted to reject rather than hang. The gated integration tests separately cover the real Pi happy path where the result is included and Stop follows.

packages/server/src/agent/agentSessionManager.ts
rsolmano: Confirmed intentional and documented in `a0858fd9`. Workspace archival is destructive teardown: it aborts unanswered questions and deletes their transcripts, so preserving a restart-answerable dialog would be misleading. Graceful application shutdown remains the only preservation path.

OLavrik (approved):

packages/server/src/agent/agentSessionManager.ts
```diff
@@ -1233,7 +1290,13 @@ export async function abortSession(
 	restoreQueue = false,
 ): Promise<SessionQueueContent | undefined> {
 	const entry = mustGetEntry(sessionId);
-	const restoredQueue = restoreQueue ? clearQueueSession(sessionId) : undefined;
+	let restoredQueue = restoreQueue ? clearQueueSession(sessionId) : undefined;
+	const acceptedResult = entry.askUserQuestionWaiters.prepareAbort();
+	if (acceptedResult) await acceptedResult.catch(() => {});
```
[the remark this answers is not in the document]
rsolmano: Fixed in `ca313117`. Stop now gives accepted-result persistence a 1s grace period, then signals Pi abort. The waiter snapshots and verifies the complete native result, and a real stalled `tool_result` regression proves Stop settles while a replaced result rejects the answer RPC.

